PR #51, opened 2023-12-04
Enhanced Deployment 

lukasfrank:
# Proposed Changes

- Added `onmetal` deployment artefacts 
- Added `ironcore-dev` deployment artefacts 
- Added init container to prepare host


config/libvirt-provider-onmetal/libvirt-provider-rbac/role.yaml
```diff
@@ -0,0 +1,161 @@
...
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - compute.api.onmetal.de
+  resources:
+  - machines
+  verbs:
+  - create
+  - delete
```
hardikdr, line 25: Does it need access to delete all the resources(Machines, etc) included in this Role?

hardikdr (commented):
Thanks @lukasfrank

config/libvirt-provider-onmetal/libvirt-provider-rbac/role.yaml
lukasfrank: For most of the resources it's probably not needed but I would keep it for now in sync with the broker roles.

hardikdr (approved):

PR #65, opened 2023-12-11
Support Machine Exec

### LINKED ISSUE #8: Implement the Exec Interface for Machines
# Summary

This is an issue about implementing the `Exec` feature for the Machines, in order to offer a similar experience to the users as when `kubectl exec -it pod_name` is used. 

- The high-level proposal for Exec implementation for ironcore can be found here: https://github.com/ironcore-dev/ironcore/blob/main/docs/proposals/02-machine-console-access.md
- The overall feature also requires an ironcore plugin for the `kubectl` , which can be found here: https://github.com/onmetal/kubectl-onmetal. (to be updated)
 
## Basic example

This feature should mainly implement the IRI gRPC interface `Exec` within the libvirt-provider:

```golang
func (s *Server) Exec(ctx context.Context, req *ori.ExecRequest) (*ori.ExecResponse, error) {
	return &ori.ExecResponse{}, nil
}
```

## Motivation

Better end-user experience, and better-aligned approach to console-access with K8s primitives.

hardikdr:
# Proposed Changes
This PR implements the Exec interface from the IRI MachineRuntime Interface. This is achieved by creating a streaming http server that is responsible for upgrading the connection to websocket and responding to the Exec API calls for a given VM.

- Feature can be tested with the `irictl` cli:
```
./bin/irictl-machine --address=unix:/var/run/ori-machinebroker.sock exec machine_id
```

- The StreamCreationTimeout and StreamIdleTimeout are defaulted, and open for discussion if there are arguments for better values.

Fixes #8


provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+		log.Error(err, "Error creating exec handler")
+		code := http.StatusInternalServerError
+		http.Error(w, http.StatusText(code), code)
+		return
+	}
+
+	handler.Handle(w, req, remotecommandserver.ExecOptions{})
+}
+
+func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser, resize remotecommand.TerminalSizeQueue) error {
```
so-sahu, line 84: Unused parameter `resize`.

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	// ReadInput: go routine to read the input from the reader, and write to the terminal.
+	go func() {
+		defer wg.Done()
+
+		buf := make([]byte, 1024)
+		for {
+			n, err := inputReader.Read(buf)
+			if err != nil {
+				if _, ok := err.(term.EscapeError); ok {
+					_, _ = f.Write(buf) // This is to close the writer.
```
so-sahu, line 123: error returned from Write is shadowed.

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+				log.Error(err, "error writing to the file descriptor")
+				return
+			}
+		}
+	}()
+
+	// WriteOutput: go routine for writing the output back to the Writer.
+	go func() {
+		defer wg.Done()
+		_, _ = io.Copy(out, f)
```
so-sahu, line 142: error returned from Copy is shadowed.

so-sahu (changes requested):

provider/cmd/app/app.go
```diff
@@ -60,15 +66,18 @@ func init() {
...
+
 	RootDir string
 
 	PathSupportedMachineClasses string
 
 	ApinetKubeconfig string
 
-	Libvirt   LibvirtOptions
-	NicPlugin *networkinterfaceplugin.Options
+	LibvirtOpts LibvirtOptions
```
lukasfrank, line 79: pls keep `Libvirt` because it is clear that we are talking about options and it reads much better `options.Libvirt.Socket`

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	machineID := e.ExecRequest.MachineId
+	log := logr.FromContextOrDiscard(ctx).WithName(machineID)
+
+	_, err := e.Libvirt.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(machineID))
+	if err != nil {
+		if !libvirtutils.IsErrorCode(err, libvirt.ErrNoDomain) {
+			return fmt.Errorf("error looking up domain: %w", err)
+		}
+		return fmt.Errorf("machine %s has not yet been synced", machineID)
+	}
```
lukasfrank, line 96: A more fine grained error handling would be nice. It might make sense to check in the store if: 
-  the machine is present
-  the machine is in an available state

lukasfrank (changes requested):

hardikdr:
Fyi: https://github.com/ironcore-dev/libvirt-provider/issues/71 for the future.

cc @afritzler (credits for pointing out)

provider/cmd/app/app.go
```diff
@@ -322,6 +349,11 @@ func Run(ctx context.Context, opts Options) error {
 		return runGRPCServer(ctx, setupLog, log, srv, opts)
 	})
 
+	g.Go(func() error {
+		setupLog.Info("Starting streaming server")
+		return runStreamingServer(ctx, setupLog, log, srv, opts)
```
lukas016, line 354: is error from runStreamingServer logged somewhere?

provider/cmd/app/app.go
```diff
@@ -357,3 +389,27 @@ func runGRPCServer(ctx context.Context, setupLog logr.Logger, log logr.Logger, s
 	}
 	return nil
 }
+
+func runStreamingServer(ctx context.Context, setupLog, log logr.Logger, srv *server.Server, opts Options) error {
```
lukas016, line 393: why are you using 2 loggers here?

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	wg.Add(2)
+	// ReadInput: go routine to read the input from the reader, and write to the terminal.
+	go func() {
+		defer wg.Done()
+
+		buf := make([]byte, 1024)
+		for {
+			n, err := inputReader.Read(buf)
+			if err != nil {
+				if _, ok := err.(term.EscapeError); ok {
```
lukas016, line 122: it is old way. it is recommended use errors.Is currently.

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	// WriteOutput: go routine for writing the output back to the Writer.
+	go func() {
+		defer wg.Done()
+		_, _ = io.Copy(out, f)
+		log.Info("Closed writing to the terminal")
+	}()
+
+	if err := cmd.Wait(); err != nil {
+		// Avoid returning so that the function can verify if all go routines are terminated.
+		log.Error(err, "Console command interrupted")
```
lukas016, line 148: you don't have same format of error message. You combine uppercase and lowercase first letter for messages. I suggest use one format ideally lowercase: Primary error has to be lowercases which are part of log output, so i recommend follow same format: https://google.github.io/styleguide/go/decisions.html#error-strings

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	// Wrap the input stream with an escape proxy. Escape Sequence Ctrl + ] = 29
+	inputReader := term.NewEscapeProxy(in, []byte{29})
+
+	wg.Add(2)
+	// ReadInput: go routine to read the input from the reader, and write to the terminal.
+	go func() {
+		defer wg.Done()
+
+		buf := make([]byte, 1024)
+		for {
```
lukas016, line 119: is goroutine ended with context or they can be stuck in infinity loop?

provider/server/exec.go
```diff
@@ -5,11 +5,150 @@ package server
...
+	ExecRequest     *iri.ExecRequest
+	VirshExecutable string
+}
+
+func (s *Server) Exec(ctx context.Context, req *iri.ExecRequest) (*iri.ExecResponse, error) {
+	machineID := req.MachineId
+	log := s.loggerFrom(ctx, "MachineID", machineID)
+
+	log.V(1).Info("Inserting request into cache")
+	token, err := s.execRequestCache.Insert(req)
```
lukas016, line 42: what is expiration time for it?

lukas016: 1 min, it is ok

hardikdr (commented):
@lukas016 @lukasfrank @so-sahu thanks a lot for the review, I have replied to the comments, and made a few changes. 

PS: I ended up force-pushing instead of splitting, changes are minimal though.

provider/server/exec.go
lukas016: correct way is use _ as name variable. Interface will still same but developers directly see you don't use this variable. So pls do it properly.

provider/server/exec.go
lukas016: you can add comment about it minimally.

provider/cmd/app/app.go
lukas016: I don't think. Did you read errgroup docs? because g.Wait() will return first error and you return errors on 2 places into errgroup. Did you tested it?

```text
Wait blocks until all function calls from the Go method have returned, then returns the first non-nil error (if any) from them.
```

provider/server/exec.go
lukas016: i am little confuse because error interface requires Error() method which returns string: https://pkg.go.dev/builtin#error

```golang
type error interface {
	Error() [string](https://pkg.go.dev/builtin#string)
}
```

provider/server/exec.go
hardikdr: done.

provider/cmd/app/app.go
lukas016: did you read code from lukas frank, where he logs each error for other components of code as machine events separately? So my suggestion follow standard pattern in libvirt-provider code. You cannot hide errors from other components because it can end in graceful shutdown process with really serious error and we won't know it because you will always ignore it.

lukas016: i will create issue for improve error logging.

provider/server/exec.go
lukas016: EscapError is structure type and it cannot be used with errors.Is and it require create object of specific type. We decide keep current solution because benefits isn't huge.

lukas016 (approved):

provider/server/exec.go
```diff
@@ -5,11 +5,165 @@ package server
...
+	handler.Handle(w, req, remotecommandserver.ExecOptions{})
+}
+
+func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser, _ remotecommand.TerminalSizeQueue) error {
+	var wg sync.WaitGroup
+
+	machineID := e.ExecRequest.MachineId
+	log := logr.FromContextOrDiscard(ctx).WithName(machineID)
+
+	// Check if the apiMachine doesnt exist, to avoid making the libvirt-lookup call.
```
lukasfrank, line 98:
```suggestion
	// Check if the apiMachine doesn't exist, to avoid making the libvirt-lookup call.
```

lukasfrank (changes requested):
@hardikdr Can you please sign your commits?

lukasfrank:
@hardikdr Can you also provide some basic tests at least for the server implementation?

hardikdr:
@lukasfrank Signed the commit and added the test for the Exec interface.

so-sahu (dismissed):
LGTM

provider/server/exec.go
```diff
@@ -5,11 +5,165 @@ package server
...
+	token, err := s.execRequestCache.Insert(req)
+	if err != nil {
+		return nil, err
+	}
+
+	log.V(1).Info("Returning url with token")
+	return &iri.ExecResponse{
+		Url: s.buildURL("exec", token),
+	}, nil
+}
```
lukasfrank, line 53: Can you please return a `not found error` if the machine is not in the store?

provider/server/exec.go
```diff
@@ -5,11 +5,165 @@ package server
...
+
+func (s *Server) ServeExec(w http.ResponseWriter, req *http.Request, token string) {
```
lukasfrank, line 55: The 'unhappy' path of `ServeExec` should be also testet. Once we spin up a `VM` in the pipeline we can do the entire flow.

lukasfrank (changes requested):

provider/server/exec.go
hardikdr: Sure, I added a check to verify the store, before returning the token.

provider/server/exec.go
hardikdr: The meaningful tests for `ServeExec` could be a little complex to mock, and better handled in the pipeline, considering its nature and dependencies. Nevertheless, I have added a simple test in the http-server for the flow of unknown tokens, that eventually calls ServeExec.

lukasfrank (changes requested):

provider/server/exec_test.go
```diff
@@ -0,0 +1,43 @@
...
+		machineID := resCreate.GetMachine().Metadata.Id
+
+		By("issuing exec for the test machine")
+		resExec, err := srv.Exec(ctx, &iri.ExecRequest{MachineId: machineID})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("inspecting the result")
+		u, err := url.ParseRequestURI(resExec.Url)
+		Expect(err).NotTo(HaveOccurred(), "url is invalid: %q", resExec.Url)
+		Expect(u.Host).To(Equal("localhost:8080"))
```
so-sahu, line 39: Optional: This is directly linked with the baseURL you have set in the test suite i.e. `baseURL              = "http://localhost:8080"`.
I would suggest not to hardcode the value here, as it will fail if someone changes the baseURL.

provider/server/exec_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+var _ = Describe("Exec", func() {
+
+	It("should return an exec-url with a token", func(ctx SpecContext) {
+		By("creating the test machine")
+		resCreate, err := srv.CreateMachine(context.TODO(), &iri.CreateMachineRequest{
+			Machine: &iri.Machine{
+				Metadata: &irimeta.ObjectMetadata{},
+				Spec: &iri.MachineSpec{
+					Class: "x3-xlarge",
```
so-sahu, line 24: Optional: `x3-xlarge` is used in multiple places. So, it can be set as a variable and used in all the places.

provider/server/exec_test.go
```diff
@@ -0,0 +1,43 @@
...
+	irimeta "github.com/ironcore-dev/ironcore/iri/apis/meta/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Exec", func() {
+
+	It("should return an exec-url with a token", func(ctx SpecContext) {
+		By("creating the test machine")
+		resCreate, err := srv.CreateMachine(context.TODO(), &iri.CreateMachineRequest{
```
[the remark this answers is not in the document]
hardikdr: Sure, that's a miss.

provider/server/exec_test.go
hardikdr: Sure, done, thanks for pointing.

provider/server/exec_test.go
hardikdr: Done, thanks.

provider/server/exec_test.go
hardikdr: I will fix this soon, seeing a small hiccup.

provider/server/exec_test.go
```diff
@@ -0,0 +1,42 @@
...
+		machineID := resCreate.GetMachine().Metadata.Id
+
+		By("issuing exec for the test machine")
+		resExec, err := srv.Exec(ctx, &iri.ExecRequest{MachineId: machineID})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("inspecting the result")
+		u, err := url.ParseRequestURI(resExec.Url)
+		Expect(err).NotTo(HaveOccurred(), "url is invalid: %q", resExec.Url)
+		Expect(u.Host).To(Equal(baseURL))
```
so-sahu, line 38: The baseURL has to be divided into host and schema.

hardikdr: Yup, noticed that, but had to rush for lunch. I have fixed that, can you please check again.

so-sahu (approved):
LGTM

provider/server/exec.go
```diff
@@ -5,11 +5,182 @@ package server
...
+	Libvirt         *libvirt.Libvirt
+	ExecRequest     *iri.ExecRequest
+	VirshExecutable string
+	Machine         *api.Machine
+}
+
 func (s *Server) Exec(ctx context.Context, req *iri.ExecRequest) (*iri.ExecResponse, error) {
+	machineID := req.MachineId
+	log := s.loggerFrom(ctx, "MachineID", machineID)
+
```
lukasfrank, line 46:
```suggestion
	log := s.loggerFrom(ctx, "MachineID", req.MachineId)
```

variable is used only once, pls remove it or replace others as well

lukasfrank (approved):

PR #69, opened 2023-12-12
Implemented `Volume` encryption support

lukasfrank:
# Proposed Changes

- Implemented `Volume` encryption support


hardikdr (approved):
Thanks @lukasfrank for the PR.

hardikdr:
Merging it already, considering itss dependency on the deployment that Lukas would like to test now.

Issue #48, opened 2023-11-30
Support NUMA Domain Pinning and Usage of Hugepages


hardikdr:
# Summary

In the multiprocessing architectures, each processor is attached with its own local memory, also known as NUMA domains. VirtualMachines would be greatly benefitted if larger chunks or all of its allocated memory belong to single NUMA domain, or to least number of them.

libvirt-provider currently depends on host-OS to make a decision of memory allocation, this can be changed by explicitly pinning the NUMA domains in the libvirt-domain xml. Considering the possible complexities of Memory Allocation regarding the fragmentation, this feature should be best effort for the start, and cant be guaranteed for all VMs. 

To make the NUMA affinity more effective, libvirt-provider should also identify the hugepage size of the host, and integrate it into the libvirt-domain.

It would also be nice to be able to enable or disable the NUMA domain pinning with the configurable flag.

## Basic example

This shouldn't require any changes to the API. 

Implementation:
```
➜  ~ numactl --hardware --show
available: 2 nodes (0-1)
node 0 cpus: 0 1 2 3 4 5 6 7 8 9 10 11 24 25 26 27 28 29 30 31 32 33 34 35
node 0 size: 386684 MB
node 0 free: 378067 MB
node 1 cpus: 12 13 14 15 16 17 18 19 20 21 22 23 36 37 38 39 40 41 42 43 44 45 46 47
node 1 size: 387033 MB
node 1 free: 366787 MB
```
VM should consume memory entirely from either node0, or node1, on a best effort basis.

## Motivation

Better performance for the VMs.

PR #70, opened 2023-12-12
Support NUMA Domain Pinning and Usage of Hugepages

so-sahu:
# Proposed Changes

- Adding support for NUMA domain pinning and usage of hugepages .
- Adding UTs for the same.
-  Configurable flags able to enable or disable the NUMA domain pinning and hugepages. 
- Document the algorithm for the NUMA aware placements.

### DIFF
--- a/docs/architecture/concepts/numa-algorithm.md
+++ b/docs/architecture/concepts/numa-algorithm.md
@@ -1,0 +1,84 @@
+# Algorithm for NUMA-aware placement of VirtualMachines
+
+- [Preparation Steps](#preparation-steps)
+- [Core Logic](#core-logic)
+
+## Preparation Steps
+
+1. Define an example `Machine` object with the configuration of 8G and 4 Cores and reconcile it.
+
+1. Build the `CPUTopology` map once at the start, which creates a Go-map representation of the CPU topology on the host, with `map[int][]int`.
+
+    ```text
+    Example:
+    CPUTopology {
+    node-id-0: [0 1 2 3 4 5 6 7 8 9 10 11 24 25 26 27 28 29 30 31 32 33 34 35],
+    node-id-1: [12 13 14 15 16 17 18 19 20 21 22 23 36 37 38 39 40 41 42 43 44 45 46 47],
+    }
+    ```
+
+1. In every new `Machine` reconciliation, analyze the host machine.
+
+    - Check and store the default huge page size.
+    - Check if 8G is available in total on the host machine to be allocated in terms of pre-allocated hugepages.
+
+1. Create the map called `pins`, which represents the number of VMs assigned to each of the CPU cores, sorted from least to highest.
+
+    ```text
+    Example: pins = [0:1 1:2 2:2 ... 33:3 34:3 35:4]
+    ```
+
+1. Create a sorted map from `CPUTopology`, which is an array of CPU cores per NUMA zones, sorted from least to highest number of VMs assigned to the CPU cores.
+
+    ```text
+    Example:
+    Sorted CPUTopology {
+    node-id-0: [0 1 2 3 4 5 6 7 8 9 10 11 24 25 26 27 28 29 30 31 32 33 34 35],
+    node-id-1: [12 13 14 15 16 17 18 19 20 21 22 23 36 37 38 39 40 41 42 43 44 45 46 47],
+    }
+    ```
+
+## Core Logic
+
+1. Choose the nodeset with the most available Memory, either 0 or 1.
+
+    - If neither of the nodesets have enough memory available, allocate the remaining memory from both nodesets.
+
+    ```text
+    Example: AllocatedMemory - [0:6G 1:2G]
+    ```
+
+    - Based on the proportion of memory, determine the number of CPUs to choose from the CPUTopology from both NodeSets.
+
+    ```text
+    Example:
+    If the proportion of memory is 75% from NodeSet0 and 25% from NodeSet1, and there are 4 CPUs available in total, then 3 CPUs will be allocated from NodeSet0 and 1 CPU will be allocated from NodeSet1. Choose ceiling values in the case of non-integer values.
+    ```
+
+2. Create the domain XML from the information above.
+
+    ```xml
+    <memoryBacking>
+        <hugepages/>
+    </memoryBacking>
+    <vcpu placement='static'>4</vcpu>
+    <cputune>
+        <vcpupin vcpu='0' cpuset='0'/>
+        <vcpupin vcpu='1' cpuset='1'/>
+        <vcpupin vcpu='2' cpuset='2'/>
+        <vcpupin vcpu='3' cpuset='15'/>
+    </cputune>
+
+    <numatune>
+        <memory mode='strict' nodeset='0-   1'/>
+        <memnode cellid='0' mode='strict' nodeset='0'/>
+        <memnode cellid='1' mode='strict' nodeset='1'/>
+    </numatune>
+
+    <cpu mode='custom' match='exact' check='full'>
+        <numa>
+        <cell id='0' cpus='0-2' memory='6G' unit='KiB'/>
+        <cell id='1' cpus='3' memory='2G' unit='KiB'/>
+        </numa>
+    </cpu>
+    ```
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -52,9 +52,12 @@ var (
 	}
 )
 
+type TuneFunc func(context.Context, *libvirtxml.Domain, *api.Machine, *MachineReconciler) error
+
 type MachineReconcilerOptions struct {
 	GuestCapabilities      guest.Capabilities
 	TCMallocLibPath        string
+	DomainTuners           []TuneFunc
 	ImageCache             virtletimage.Cache
 	Raw                    raw.Raw
 	Host                   virtlethost.Host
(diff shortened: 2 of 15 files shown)


docs/architecture/concepts/numa-algorithm.md
```diff
@@ -0,0 +1,84 @@
...
+    ```
+
+    - Based on the proportion of memory, determine the number of CPUs to choose from the CPUTopology from both NodeSets.
+
+    ```text
+    Example:
+    If the proportion of memory is 75% from NodeSet0 and 25% from NodeSet1, and there are 4 CPUs available in total, then 3 CPUs will be allocated from NodeSet0 and 1 CPU will be allocated from NodeSet1. Choose ceiling values in the case of non-integer values.
+    ```
+
+2. Create the domain XML from the information above.
```
lukas016, line 58: optional: use automatically numbered list (1.) as in previous list

pkg/host/cpu.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"encoding/xml"
+	"fmt"
+
+	"github.com/digitalocean/go-libvirt"
+	"libvirt.org/go/libvirtxml"
+)
+
+// CpuTopology return a list of related to NUMA node CPU cores
+func CpuTopology(lv *libvirt.Libvirt) (map[int][]int, error) {
```
lukas016, line 15: recommendation: Acronyms should be all capitals. It is golang convention

pkg/host/hugepages.go
```diff
@@ -0,0 +1,115 @@
...
+func InitHugepageTuner(lv *libvirt.Libvirt, HugePageSize uint64) (controllers.TuneFunc, error) {
+	nt := HugepageTuner{
+		Lv:           lv,
+		HugePageSize: HugePageSize,
+	}
+	return nt.Tune, nil
+}
+
+func (n *HugepageTuner) Tune(ctx context.Context, domain *libvirtxml.Domain, _ *api.Machine, _ *controllers.MachineReconciler) error {
+	log, err := logr.FromContext(ctx)
```
lukas016, line 32: optional: FromContextOrDiscard can be safer

pkg/host/cpu.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// getVcpuPinInfo return the CPU affinity setting of all virtual CPUs of domain
+func getVcpuPinInfo(lv *libvirt.Libvirt, domain libvirt.Domain) ([][]bool, error) {
+	_, _, _, _, rNodes, rSockets, rCores, rThreads, err := lv.NodeGetInfo()
+	if err != nil {
+		return nil, err
+	}
+	npcpus := rNodes * rSockets * rCores * rThreads
+	maplen := (npcpus + 7) / 8
```
lukas016, line 48: why 7 / 8? what are these numbers?

pkg/host/hugepages.go
```diff
@@ -0,0 +1,115 @@
...
+
+	domain.MemoryBacking = &libvirtxml.DomainMemoryBacking{
+		MemoryHugePages: &libvirtxml.DomainMemoryHugepages{},
+	}
+
+	return nil
+}
+
+func (n *HugepageTuner) isEnoughFreePagesAllocated(memory uint64) (bool, error) {
+	needMemoryKb, _ := AlignMemory(memory, 1024)
```
lukas016, line 62: optional: you use 1024 on multiple places. it will be better set as varialbe or  move into constants

pkg/host/numa.go
```diff
@@ -0,0 +1,313 @@
...
+	for node, cpus := range nodeCPUs {
+		//	CPU cores with less pinned vm count go first
+		sort.SliceStable(cpus, func(i, j int) bool {
+			return pins[cpus[i]] < pins[cpus[j]]
+		})
+		nodeCPUs[node] = cpus
+	}
+
+	var totalMemory uint64
+	nodeIDs := make([]int, 0, len(allocatableMemoryPerNode))
```
lukas016, line 114: optional: you call this len(allocatableMemoryPerNode) 2 times. You can store value into variable and use it on line 127

pkg/controllers/machine_controller.go
```diff
@@ -104,6 +108,7 @@ type MachineReconciler struct {
 	libvirt           *libvirt.Libvirt
 	guestCapabilities guest.Capabilities
 	tcMallocLibPath   string
+	DomainTuners      []TuneFunc
```
lukas016, line 111: @lukasfrank is reconciler running in multiple goroutines in libvirt-provider? Because virtlet runs with 1 worker in time. This concept can create unexpected behavior with multiple goroutines.

provider/cmd/app/app.go
```diff
@@ -80,6 +81,10 @@ type LibvirtOptions struct {
 	PreferredMachineTypes []string
 
 	Qcow2Type string
+
+	DisableNuma      bool
```
lukas016, line 85: are their libvirt options or libvirt-provider options?

lukas016 (changes requested):

pkg/host/cpu.go
so-sahu: It is a common idiom to calculate the byte length needed to represent a bitmap or array where each bit corresponds to a specific CPU, ensuring that there is enough space for rounding

provider/cmd/app/app.go
so-sahu: I believe this section of libvirt-provider has options to interact with libvirt.

pkg/controllers/machine_controller.go
lukas016: we discussed it on sync call. libvirt-provider can process 10 machine at the same time and call tune functions in the same time. It can create unexpected behavior because each tuna function calculate resources independently.

so-sahu: So, is it like we need to make the resource calculation thread safe ?

pkg/host/cpu.go
lukas016: 8 is count of bits for byte
what is 7?

pkg/controllers/machine_controller.go
lukas016: it depends how do you mean it but technically yes. Primary problem is missing central management for pinning vcpu.

provider/cmd/app/app.go
lukas016: yes but we interact with libvirt on multiple levels. Primary we have libvirt client and settings specific for libvirt daemon. After we have this tuna functions which modified libvirt domain but they are independent on client or daemon. If you look on fields, you have multiple different fields in one structure which serve for different purpose and they are use in different places of code. So i prefer create for tunning separate structure. It will improve readability of code and you will very nice see which fields are specific for tunning only.

so-sahu: Would you like to suggest any name for the structure.

pkg/controllers/machine_controller.go
so-sahu: Central management will be taken care under the the Resource manager umbrella I believe. So, what kind of modifications are you expecting in this PR.

pkg/host/cpu.go
so-sahu: it's to ensure that there is enough space for rounding.

pkg/controllers/machine_controller.go
lukas016: i am not sure about it. pinning cpu isn't resource. It is some state, how vms are assign to cpu cores, what isn't important during scheduling. But it is for long discussion.
But minimally, if we don't want to improve code in this PR, it will ideally change worker count to 1 for minimize impact of this problem.

provider/cmd/app/app.go
lukas016: Tunning, Tune, ... up to you

pkg/host/numa.go
```diff
@@ -0,0 +1,313 @@
...
+	domain.CPU.Numa = numaCfg
+
+	if domain.CPUTune == nil {
+		domain.CPUTune = &libvirtxml.DomainCPUTune{}
+	}
+
+	domain.CPUTune.VCPUPin = vcpuPins
+
+	return nil
+}
```
lukasfrank, line 202: `Tune` is called concurrently. At the moment the calculation can be wrong therefore.

pkg/host/hugepages.go
```diff
@@ -0,0 +1,115 @@
...
+
+func InitHugepageTuner(lv *libvirt.Libvirt, HugePageSize uint64) (controllers.TuneFunc, error) {
+	nt := HugepageTuner{
+		Lv:           lv,
+		HugePageSize: HugePageSize,
+	}
+	return nt.Tune, nil
+}
+
+func (n *HugepageTuner) Tune(ctx context.Context, domain *libvirtxml.Domain, _ *api.Machine, _ *controllers.MachineReconciler) error {
```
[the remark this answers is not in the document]
lukasfrank: Same here, called concurrently

lukasfrank (changes requested):

pkg/host/numa.go
lukas016: it is solving in other thread

hardikdr:
Added `do-not-merge` label.

Potential issues were identified related to concurrent resource access by different worker threads. This could impact NUMA-pinning and memory allocation, which are particularly also crucial for SGX support later.

To address these concerns, a more detailed design is to be discussed as a follow-up. This design should comprehensively cover NUMA-pinning, hugepage allocation, thread safety, and SGX support, ensuring efficient resource management.

lukasfrank:
Closing this PR for now, since it needs more considerations. Design proposal will follow in  #48.

PR #84, opened 2023-12-20
Configuring Metadata in DomainXML

### LINKED ISSUE #82: Configure Metadata in DomainXML
# Summary

The current implementation of the libvirt-provider lacks the necessary metadata configuration in the generated domainXML.

## Motivation

The metadata can include essential information about the domain, influencing its behavior, configuration, and interactions with other components.

In an ideal libvirt-provider scenario, the metadata structure should resemble the following:

```
<metadata>
    <libvirtprovider:metadata xmlns:libvirtprovider="https://github.com/ironcore-dev/libvirt-provider">
          <libvirtprovider:namespace>default</libvirtprovider:namespace>
          <libvirtprovider:name>machine-hd1</libvirtprovider:name>
    </libvirtprovider:metadata>
</metadata>
```

so-sahu:
# Proposed Changes

- Adding api achine metadata from iri machine
- Configuring metadata in domainXML from api machine

Fixes #82


pkg/controllers/machine_controller.go
```diff
@@ -554,6 +561,21 @@ func (r *MachineReconciler) domainFor(
 	return domainDesc, volumeStates, nicStates, nil
 }
 
+func (r *MachineReconciler) setDomainMetadata(machine *api.Machine, domain *libvirtxml.Domain) error {
+	domainMetadata := &providermeta.LibvirtProviderMetadata{
+		Namespace: machine.Metadata.Labels[ironcorev1alpha1.MachineNamespaceLabel],
```
lukas016, line 566: what will be result, if namespace isn't defined in labels?

provider/apiutils/apiutils.go
```diff
@@ -109,6 +110,14 @@ func GetClassLabel(o api.Object) (string, bool) {
 	return class, found
 }
 
+func SetMachineNamespaceLabel(o api.Object, namespace string) {
```
lukas016, line 113: do you need this functions? It will save only one argument from 3. I don't see big benefits here or extending of SetLabel logic

provider/apiutils/apiutils.go
```diff
@@ -109,6 +110,14 @@ func GetClassLabel(o api.Object) (string, bool) {
...
+	metautils.SetLabel(o, ironcorev1alpha1.MachineNamespaceLabel, namespace)
+}
+
+func SetMachineNameLabel(o api.Object, namespace string) {
+	metautils.SetLabel(o, ironcorev1alpha1.MachineNameLabel, namespace)
```
lukas016, line 118: are you sure with namespace variable?

lukas016 (changes requested):

provider/apiutils/apiutils.go
so-sahu: I will change it to `name`. Thanks for pointing.

provider/apiutils/apiutils.go
so-sahu: I am trying to follow the existing conversion with `SetClassLabel` and `SetManagerLabel` being already there.

pkg/controllers/machine_controller.go
so-sahu: It will be empty.

lukas016: yes i know, but are you okay with that? because map return 2 values (value for key and boolean if key exists)

provider/apiutils/apiutils.go
so-sahu: Changed.

pkg/controllers/machine_controller.go
lukas016: we had call. if label is missing, it isn't problem of this part of code

lukas016 (dismissed):

provider/server/machine_create.go
```diff
@@ -78,6 +79,8 @@ func (s *Server) createMachineFromIRIMachine(ctx context.Context, log logr.Logge
 		return nil, fmt.Errorf("failed to set metadata: %w", err)
 	}
 	apiutils.SetClassLabel(machine, iriMachine.Spec.Class)
+	apiutils.SetMachineNamespaceLabel(machine, iriMachine.Metadata.Labels[ironcorev1alpha1.MachineNamespaceLabel])
+	apiutils.SetMachineNameLabel(machine, iriMachine.Metadata.Labels[ironcorev1alpha1.MachineNameLabel])
```
lukasfrank, line 83: If either `iriMachine.Metadata` or `iriMachine.Metadata.Labels` is empty we run in an nil-pointer exception

lukasfrank (changes requested):

provider/server/machine_create.go
lukas016: can we solve it somewhere else? because we will have problem with it always. It can be solve as setter for metedata type which can handle this nil references.

lukasfrank: pls remove since the metadata is already stored here: 
```
	if err := apiutils.SetObjectMetadata(machine, iriMachine.Metadata); err != nil {
		return nil, fmt.Errorf("failed to set metadata: %w", err)
	}
```

so-sahu: I acknowledge that if `iriMachine.Metadata` is empty, it could lead to a nil-pointer exception since Metadata is a pointer to `v1alpha1.ObjectMetadata`. To address this, I will include a condition to ensure that Metadata is not nil before proceeding.

On the other hand, `iriMachine.Metadata.Labels` is of type `map[string]string` and will be initialized to its zero value if left empty. Therefore, there is no need to be concerned about nil values. I have tested this scenario, and in such cases, the metadata in domainXML appears as follows:
```
<metadata>
    <libvirtprovider:metadata xmlns:libvirtprovider="https://github.com/ironcore-dev/libvirt-provider">
        <libvirtprovider:namespace/>
        <libvirtprovider:name/>
    </libvirtprovider:metadata>
</metadata>
```
This output demonstrates that even when`iriMachine.Metadata.Labels` is empty, the XML representation is well-formed and contains the expected structure.

Please let me know if, this will address your concern.

lukasfrank: I would propose to store the data only once and fetch it in the controller directly at the correct place.

so-sahu: ok I got your point, we can extract it from the value of annotation `libvirt-provider.ironcore.dev/labels`. Correct?

However, still your concern remains that if  `iriMachine.Metadata` is empty, then `if err := apiutils.SetObjectMetadata(machine, iriMachine.Metadata); err != nil {` will lead to nil-pointer exception.
Also in some places before using `iriMachine.Spec` the nil pointer is not bein checked, even that can be problamatic.

pkg/controllers/machine_controller.go
```diff
@@ -554,6 +561,21 @@ func (r *MachineReconciler) domainFor(
...
+	}
+	domainMetadataXML, err := xml.Marshal(domainMetadata)
+	if err != nil {
+		return err
+	}
+	domain.Metadata = &libvirtxml.DomainMetadata{
+		XML: string(domainMetadataXML),
+	}
+	return nil
+}
```
lukasfrank, line 577: Can we here maybe encode the entire labels we receive from `machine.Metadata.Labels`? 
I think a generic xml encoder for map[string]string would be nice in order to store everything (indeed from the label name)

lukasfrank: At SAP we need this information since it contains the top most cluster info as well as the info from the poollet.
It looks like this: 
```
          "downward-api.machinepoollet.api.onmetal.de/root-machine-name": "machine",
          "downward-api.machinepoollet.api.onmetal.de/root-machine-namespace": "default",
          "downward-api.machinepoollet.api.onmetal.de/root-machine-uid": "0276f002-a4d4-4b84-a1b4-94dfaa5475bb",
          "machinepoollet.api.onmetal.de/machine-name": "749112270492dadb02e5313b447429dcc6b56b72ce072f112a44961e49c61eb",
          "machinepoollet.api.onmetal.de/machine-namespace": "machinepoollet-system",
          "machinepoollet.api.onmetal.de/machine-uid": "151e0229-f3c3-456c-bf32-65d67dfb6096"
```

lukasfrank (changes requested):

provider/server/machine_create.go
lukas016: @so-sahu be careful. map is reference type and zero value for map is nil: https://go.dev/tour/moretypes/19

so-sahu: @lukas016  In Go, a nil map is a valid value for a map type, and you can perform read operations on it without causing a panic. If you try to access a key in a nil map, the result will be the zero value for the map's value type.
So, here, it's not a matter of concern.

lukas016: yes, you have true but write is problematic so pls be careful with zero values.

pkg/controllers/machine_controller.go
```diff
@@ -554,6 +562,30 @@ func (r *MachineReconciler) domainFor(
 	return domainDesc, volumeStates, nicStates, nil
 }
 
+func (r *MachineReconciler) setDomainMetadata(machine *api.Machine, domain *libvirtxml.Domain) error {
+	var irimachineLabels map[string]string
+
+	err := json.Unmarshal([]byte(machine.Metadata.Annotations[machinev1alpha1.LabelsAnnotation]), &irimachineLabels)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling iri machine labels: %w", err)
+	}
```
lukasfrank, line 571:
```suggestion
	labels, found := machine.Metadata.Annotations[machinev1alpha1.LabelsAnnotation]
	if !found {
		return nil
	}

	var irimachineLabels map[string]string
	err := json.Unmarshal([]byte(labels), &irimachineLabels)
	if err != nil {
		return fmt.Errorf("error unmarshalling iri machine labels: %w", err)
	}
```

pkg/controllers/machine_controller.go
```diff
@@ -554,6 +562,34 @@ func (r *MachineReconciler) domainFor(
 	return domainDesc, volumeStates, nicStates, nil
 }
 
+func (r *MachineReconciler) setDomainMetadata(log logr.Logger, machine *api.Machine, domain *libvirtxml.Domain) error {
+	labels, found := machine.Metadata.Annotations[machinev1alpha1.LabelsAnnotation]
+	if !found {
+		log.V(1).Info("IRI machine labels are not annotated in the API machine", "Machine", machine.ID)
```
lukasfrank, line 568: pls remove the machine id, it is already set at the start of the reconcile: 
```
	log = log.WithValues("machineId", id)
```

lukasfrank (approved):

PR #86, opened 2023-12-21
Cleanup of unused code

lukasfrank:
# Proposed Changes

- Removed not used code


hardikdr (approved):
Approving with the pov that, we can always bring back the GPU-related code when we decide to support the feature again, better designed this time.

PR #92, opened 2023-12-22
Refactor RBAC for provider and poollet

afritzler:
# Proposed Changes

Remove the provider RBAC as the provider does not need any k8s specific RBAC rules. The manager ServiceAccount is now using the generic RBAC configuration located under `config/rbac`.

Remove old kustomizations and move everything top level under `config`.


lukasfrank (changes requested):
Can you please cleanup also the `onmetal` resources?

```
config/libvirt-provider-onmetal
```

lukasfrank (approved):

so-sahu (approved):

PR #97, opened 2024-01-02
improve logging settings

### LINKED ISSUE #85: improve app logging
# Summary

We need improve and standardize format of logs records because current implementation has few issues which can do problem with integration of log collector:

- "Main error" reason for terminate of application is printed with standard fmt.Println() when other errors use zap logger.
   https://github.com/ironcore-dev/libvirt-provider/blob/main/provider/cmd/main.go#L18
- Duplicated "main error" logs:
   ```text
   Error: failed to initialize machine class registry: unable to unmarshal machine classes: error converting YAML to JSON: yaml: line    2: did not find expected alphabetic or numeric character
   failed to initialize machine class registry: unable to unmarshal machine classes: error converting YAML to JSON: yaml: line 2: did     not find expected alphabetic or numeric character
   ```
- We should use production configuration as default. So i recommend disable development level for zap logger as default settings because if you need it, you can use argument --zap-devel
- Disable print of usage for every "main error" which terminated application. We can print usage, if users use wrong argument or required arguments are missing, but it doesn't make sense print usage for application runtime error.

## Motivation

Easier integration log collectors with our application

lukas016:
* add flushing of buffers
* disable print of usage for errors
* set production configuration as default for logger
* unified format of printing error

Fixes #85

### DIFF
--- a/provider/cmd/app/app.go
+++ b/provider/cmd/app/app.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 
 	"github.com/go-logr/logr"
+	"github.com/go-logr/zapr"
 	"github.com/ironcore-dev/ironcore-image/oci/remote"
 	apinetv1alpha1 "github.com/ironcore-dev/ironcore-net/api/core/v1alpha1"
 	"github.com/ironcore-dev/ironcore/broker/common"
@@ -36,6 +37,7 @@ import (
 	"github.com/ironcore-dev/libvirt-provider/pkg/qcow2"
 	"github.com/ironcore-dev/libvirt-provider/pkg/raw"
 	"github.com/ironcore-dev/libvirt-provider/pkg/utils"
+	originzap "go.uber.org/zap"
 
 	providerhttp "github.com/ironcore-dev/libvirt-provider/provider/http"
 	"github.com/ironcore-dev/libvirt-provider/provider/networkinterfaceplugin"
@@ -125,22 +127,27 @@ func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
 	_ = cmd.MarkFlagRequired("supported-machine-classes")
 }
 
-func Command() *cobra.Command {
+// Command creates cobra command
+// It return pointer to pointer for zap logger because flags and logger are initialized in Execute function
+func Command() (*cobra.Command, **originzap.Logger) {
 	var (
-		zapOpts = zap.Options{Development: true}
+		zapOpts = zap.Options{Development: false}
 		opts    Options
 	)
 
+	var logger *originzap.Logger
 	cmd := &cobra.Command{
 		Use: "libvirt-provider",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			logger := zap.New(zap.UseFlagOptions(&zapOpts))
-			ctrl.SetLogger(logger)
+			logger = zap.NewRaw(zap.UseFlagOptions(&zapOpts))
+			ctrl.SetLogger(zapr.NewLogger(logger))
 			cmd.SetContext(ctrl.LoggerInto(cmd.Context(), ctrl.Log))
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return Run(cmd.Context(), opts)
 		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
 	}
 
 	goFlags := goflag.NewFlagSet("", 0)
@@ -150,7 +157,7 @@ func Command() *cobra.Command {
 	opts.AddFlags(cmd.Flags())
 	opts.MarkFlagsRequired(cmd)
 
-	return cmd
+	return cmd, &logger
 }
 
 func Run(ctx context.Context, opts Options) error {
--- a/provider/cmd/main.go
+++ b/provider/cmd/main.go
@@ -4,18 +4,45 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"syscall"
 
 	"github.com/ironcore-dev/libvirt-provider/provider/cmd/app"
+	originzap "go.uber.org/zap"
 	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	errTerminatedApp = "app terminated with error"
 )
 
 func main() {
 	ctx := ctrl.SetupSignalHandler()
 
-	if err := app.Command().ExecuteContext(ctx); err != nil {
-		fmt.Println(err.Error())
-		os.Exit(1)
+	cmd, ptrLogger := app.Command()
+	exitCode := int(0)
+	if err := cmd.ExecuteContext(ctx); err != nil {
+		exitCode = 1
+
+		// logger can be nil, if flags weren't parsed
+		if *ptrLogger != nil {
+			(*ptrLogger).Error(errTerminatedApp, originzap.NamedError("error", err))
+		} else {
+			fmt.Fprintln(os.Stderr, errTerminatedApp+": "+err.Error())
+		}
 	}
+
+	if *ptrLogger != nil {
+		err := (*ptrLogger).Sync()
+		// https://github.com/uber-go/zap/issues/991
+		if err != nil && !(errors.Is(err, syscall.ENOTTY) || errors.Is(err, syscall.EINVAL)) {
+			// it cannot be printed with logger, this error will extremely rare hopefully
+			fmt.Fprintln(os.Stderr, "flushing of logger buffer wasn't successfully:", err)
+		}
+	}
+
+	os.Exit(exitCode)
+
 }


provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
-	if err := app.Command().ExecuteContext(ctx); err != nil {
-		fmt.Println(err.Error())
-		os.Exit(1)
+	cmd, ptrLogger := app.Command()
+	exitCode := int(0)
+	if err := cmd.ExecuteContext(ctx); err != nil {
+		exitCode = 1
+
+		// logger can be nil, if flags weren't parsed
+		if *ptrLogger != nil {
```
so-sahu, line 30: `ptrLogger != nil` check can be added for safety to ensure that the pointer to the logger is not nil before attempting to dereference it.

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+			(*ptrLogger).Error(errTerminatedApp, originzap.NamedError("error", err))
+		} else {
+			fmt.Fprintln(os.Stderr, errTerminatedApp+": "+err.Error())
+		}
 	}
+
+	if *ptrLogger != nil {
```
so-sahu, line 37: `ptrLogger != nil` check can be added for safety to ensure that the pointer to the logger is not nil before attempting to dereference it.

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+		err := (*ptrLogger).Sync()
+		// https://github.com/uber-go/zap/issues/991
+		if err != nil && !(errors.Is(err, syscall.ENOTTY) || errors.Is(err, syscall.EINVAL)) {
+			// it cannot be printed with logger, this error will extremely rare hopefully
+			fmt.Fprintln(os.Stderr, "flushing of logger buffer wasn't successfully:", err)
```
so-sahu, line 42: typo: `successful`

provider/cmd/app/app.go
```diff
@@ -125,22 +127,27 @@ func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
 	_ = cmd.MarkFlagRequired("supported-machine-classes")
 }
 
-func Command() *cobra.Command {
+// Command creates cobra command
+// It return pointer to pointer for zap logger because flags and logger are initialized in Execute function
+func Command() (*cobra.Command, **originzap.Logger) {
```
so-sahu, line 132: Double references can be avoided for code simplicity.

so-sahu (changes requested):

provider/cmd/main.go
lukas016: it isn't technically possible with current code because i always will return pointer to variable

provider/cmd/app/app.go
lukas016: how?

provider/cmd/app/app.go
```diff
@@ -125,22 +127,27 @@ func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
...
 	var (
-		zapOpts = zap.Options{Development: true}
```
afritzler, line 130: I would control this via a command line flag.

provider/cmd/app/app.go
```diff
@@ -125,22 +127,27 @@ func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
...
 		opts    Options
 	)
 
+	var logger *originzap.Logger
 	cmd := &cobra.Command{
 		Use: "libvirt-provider",
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			logger := zap.New(zap.UseFlagOptions(&zapOpts))
-			ctrl.SetLogger(logger)
+			logger = zap.NewRaw(zap.UseFlagOptions(&zapOpts))
```
afritzler, line 142: What is the motivation behind using the Raw logger here?

provider/cmd/app/app.go
afritzler: Something like this 
```
    var devLogging bool
    flag.BoolVar(&devLogging, "dev-logging", false, "Enable development logging")
    flag.Parse()

    // Setup the logger
    opts := zap.Options{
        Development: devLogging,
    }
    logger := zap.New(zap.UseFlagOptions(&opts))

    // Set the controller-runtime logger to the created logger
    ctrl.SetLogger(logger)
```

provider/cmd/app/app.go
```diff
@@ -125,22 +127,27 @@ func (o *Options) MarkFlagsRequired(cmd *cobra.Command) {
...
+			ctrl.SetLogger(zapr.NewLogger(logger))
 			cmd.SetContext(ctrl.LoggerInto(cmd.Context(), ctrl.Log))
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return Run(cmd.Context(), opts)
 		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
```
afritzler, line 150: What is the reason to setting those to `true`?

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+	errTerminatedApp = "app terminated with error"
+)
+
 func main() {
 	ctx := ctrl.SetupSignalHandler()
 
-	if err := app.Command().ExecuteContext(ctx); err != nil {
-		fmt.Println(err.Error())
-		os.Exit(1)
+	cmd, ptrLogger := app.Command()
```
afritzler, line 24: Instead of returning a logger from the `app.Command` I would rather instantiate a `setupLog` here similar to what the controller-runtime is doing (but without importing `controller-runtime`). As this logger is only used for panic/exit scenarios it is ok to just use it during app startup. 

/ref https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/crd/main.go#L37

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+		if *ptrLogger != nil {
+			(*ptrLogger).Error(errTerminatedApp, originzap.NamedError("error", err))
+		} else {
+			fmt.Fprintln(os.Stderr, errTerminatedApp+": "+err.Error())
+		}
 	}
+
+	if *ptrLogger != nil {
+		err := (*ptrLogger).Sync()
+		// https://github.com/uber-go/zap/issues/991
```
afritzler, line 39: This issue is related to running zap under Windows. I can't think of a scenario where we would want to do that in production. In a local dev setup that should be up to the local machine setup and we should not pollute the code base with those workarounds.

afritzler (changes requested):

provider/cmd/app/app.go
lukas016: flushing of buffers

provider/cmd/app/app.go
lukas016: you can control it with flags (--zap-devel). Flags are bind on bottom of function: https://github.com/ironcore-dev/libvirt-provider/pull/97/files#diff-7c33599ec2dad35992d581529a86b575fe0e5d000ad8d558bdd3afe0688fa423R154

it is only default settings which can be override with flags.

provider/cmd/app/app.go
afritzler: Maybe instead of using `NewRaw` we should look at those (took if from the code docs):

```
// New constructs a new Logger from the provided zapcore.Core and Options. If
// the passed zapcore.Core is nil, it falls back to using a no-op
// implementation.
//
// This is the most flexible way to construct a Logger, but also the most
// verbose. For typical use cases, the highly-opinionated presets
// (NewProduction, NewDevelopment, and NewExample) or the Config struct are
// more convenient.
```

provider/cmd/main.go
lukas016: you have similar issue on mac, this is reason why i added handling of this specific errors.

provider/cmd/app/app.go
lukas016: zap library isn't real zap library but it is controller-runtime zap library which add abstraction for zap. Primary differences is support of flags which are bind later. This isn't possible do with origin zap without define own set of flags. https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/log/zap/zap.go#L229

provider/cmd/main.go
lukas016: you will have problem with inconsistent format. Because i used logger from command which can be configure with flags and you little hide implementation detail because default controller-runtime logger is null logger (dummy logger without printing message). So you can create edge cases when error messages during start won't be printed because logger wasn't initialize properly.

If you want to avoid double reference i can do retype from controller-runtime logr.Sink or we can discuss using our own wrapper library which i sent you before.

provider/cmd/app/app.go
lukas016: technically before it was running default in development mode and you had to put --zap-devel=false for set production mode. I switched logic only, so production mode is default now and you can enable devel mode with --zap-devel.

provider/cmd/main.go
afritzler: I get your point, but the setup logger is only a one of thing used during startup. If something goes sour your should fire an alert anyways. Also I have not experienced missing logs for startup failures on our side (e.g. when you set a non existing flag or some other issue occurs). I would really like to keep the start up code clean and avoid a complex pointer setup.

provider/cmd/app/app.go
afritzler: Well the we didn't really had a problem here. I would expect the code being Dev setup first and Prod setup second via config.

lukas016: Ok i can switch it back

afritzler: The default behaviour should be the most simplest setup ever: clone the repo and run the code on your box. Prod setup config is highly custom and we should not make any assumptions here.

provider/cmd/main.go
lukas016: And what flushing of buffers?

afritzler: I don't know why we need to do that. Projects like https://github.com/kubeflow/training-operator/blob/master/cmd/training-operator.v1/main.go#L104-L107 which have 100s of contribs and 1k stars don't do that. Not sure why we need this kind of special buffer flushing?

lukas016: we are using zap logging library which specifically required that: https://github.com/uber-go/zap#quick-start

https://pkg.go.dev/go.uber.org/zap#Logger.Sync

if other ignore that, it is their problem and this problem is extremely hard find out because it depends  on multiple factors.

lukas016: you don't need sync, if you are using controller-runtime zap because default output is standard error output https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/log/zap/zap.go#L173

which isn't buffered by default. But you can have problem, if you use different writer https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#WriteTo

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+)
+
 func main() {
 	ctx := ctrl.SetupSignalHandler()
 
-	if err := app.Command().ExecuteContext(ctx); err != nil {
-		fmt.Println(err.Error())
-		os.Exit(1)
+	cmd, ptrLogger := app.Command()
+	exitCode := int(0)
```
so-sahu, line 25: This exclusive conversion of 0 to int is not needed.

provider/cmd/main.go
lukas016: FYI: kubernetes uses flush for klog too https://github.com/kubernetes/component-base/blob/master/logs/logs.go#L175

provider/cmd/main.go
lukas016: it isn't retyping in this case. I specified type only. I checked assembly and both format produce same assembly code. If it is confusing for you, I can use var syntax but i prefer define specific type.

provider/cmd/main.go
```diff
@@ -4,18 +4,45 @@
...
+	if err := cmd.ExecuteContext(ctx); err != nil {
+		exitCode = 1
+
+		// logger can be nil, if flags weren't parsed
+		if *ptrLogger != nil {
+			(*ptrLogger).Error(errTerminatedApp, originzap.NamedError("error", err))
+		} else {
+			fmt.Fprintln(os.Stderr, errTerminatedApp+": "+err.Error())
```
hardikdr, line 33: Would it be possible to do something like `setupLog` here?
It's already defined in the `Run` method, should it be possible to bring it out at main, and pass it as an arg to Run then?

hardikdr (commented):
Thanks @lukas016 for the PR.

provider/cmd/main.go
lukas016: ptrLogger contains pointer to setupLog which is created later.
But if you want to simplify code, we can specified what we expect from setupLogger.
Current setupLog is created from arguments (--zap-devel, --zap-encoder, ...) which can modified format of logs. If i create setupLog in main function, i will have problem because arguments are parsed in Execute function, so it can generate different format of log records as another loggers in program. Main problem here is RunE function which return error into main and it creates chicken-egg problem. We can switch to Run without error and we will print errors directly in Run function and in this case we can simplify code.

lukas016: if you want, i can prepare second solution for comparison in different branch

lukas016: @hardikdr  i checked the code and i think ideally solution will be stop use setup logger and use unspecified logger as default and for other components we can use named loggers. Main reason is wrong usage of setup logger and named logger in Run function, their are mixed somehow randomly. Result is setup logger or using of logger name for filtration of logger messages useless. we can discuss it on sync meeting but currently it looks like you don't have any rules what and when use what creates more questions as benefits.

hardikdr: Thanks, I will check that one, and let's hold on to this one till we conclude.

PR #99, opened 2024-01-03
Add integration test workflow  and test suite for `libvirt-provider`

### LINKED ISSUE #55: Add Integration Tests
# Summary

This is an issue to track the enhancements of the Integration tests. 

## Basic example

It would be nice to have an e2e test setup similar to the ceph-provider: https://github.com/ironcore-dev/ceph-provider/blob/main/.github/workflows/integration-test.yml

## Motivation

Better test coverage, and robustness.

kasabe28:
# Proposed Changes

-  Add integration test workflow
-  Add integration test suite
-  Add integration test for CreateMachine()

Fixes #55


kasabe28:
libvirtxml.Domain Type = "qemu" and CPU = "host-passthrough" combination is not supported on ubuntu. 
Either we have to go for Type = "kvm"  and CPU = "host-passthrough" or Type = "qemu" and remove CPU (i.e. CPU=custom)
So I have added Domain type "kvm" as a workaround here
https://github.com/ironcore-dev/libvirt-provider/pull/99/files#diff-1410d3d4d059af44f7f6b6594678f460ffa2c3ba58e6444c9af1b03b51c703f3R434
Please suggest a better way if any.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,46 @@
+name: Run integration tests
+
+on:
+  push:
+    branches:
+      - enh/integration-tests
```
afritzler, line 6: This is not needed and can be removed.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    paths-ignore:
+      - 'docs/**'
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-22.04
+    #if: contains(github.event.pull_request.labels.*.name, 'integration-tests')
```
afritzler, line 15: Can you remove the commended out part here.

provider/server/machine_create_test.go
```diff
@@ -4,17 +4,20 @@
...
+	"github.com/digitalocean/go-libvirt"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
 	irimeta "github.com/ironcore-dev/ironcore/iri/apis/meta/v1alpha1"
+	libvirtutils "github.com/ironcore-dev/libvirt-provider/pkg/libvirt/utils"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("CreateMachine", func() {
+var _ = Describe("Create Machine", func() {
```
afritzler, line 15: I would keep the name in the `Describe` block the same as the method name of the runtime interface which in this case is `CreateMachine`.

provider/server/machine_list_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
+	irimeta "github.com/ironcore-dev/ironcore/iri/apis/meta/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("ListMachine", func() {
+
```
afritzler, line 14: Please remove the newline here.

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,98 @@ func TestServer(t *testing.T) {
...
+
+	userHomeDir, err := os.UserHomeDir()
+	Expect(err).NotTo(HaveOccurred())
+
+	fs := pflag.NewFlagSet("test", pflag.ContinueOnError)
+	apinetPlugin := networkinterfaceplugin.NewDefaultOptions()
+	apinetPlugin.PluginName = "apinet"
+	apinetPlugin.AddFlags(fs)
+	err = fs.Set("apinet-node-name", "test-node")
+	Expect(err).NotTo(HaveOccurred())
```
afritzler, line 85: Please run the `Set` and `Expect` in one line.

afritzler (changes requested):
Thanks a lot for this great PR! And good finding with the `kvm` option. It is great if we can use the ubuntu runners to run the integration tests. Btw. can you run a `go mod tidy` on your PR branch?

kasabe28:
> Thanks a lot for this great PR! And good finding with the `kvm` option. It is great if we can use the ubuntu runners to run the integration tests. Btw. can you run a `go mod tidy` on your PR branch?

Welcome :-). I had run `go mod tidy` on PR branch, but no change.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,42 @@
+name: Run integration tests
+
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    paths-ignore:
+      - 'docs/**'
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-22.04
```
lukas016, line 10: can we use debian or gardenlinux? it is more similar our prod env.

Makefile
```diff
@@ -175,6 +175,10 @@ addlicense: $(ADDLICENSE) ## Download addlicense locally if necessary.
 $(ADDLICENSE): $(LOCALBIN)
 	test -s $(LOCALBIN)/addlicense || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@$(ADDLICENSE_VERSION)
 
+.PHONY: integration-tests
+integration-tests:
+	CGO=1 go run github.com/onsi/ginkgo/v2/ginkgo --label-filter="integration" ./...
```
lukas016, line 180: why CGO=1?

Makefile
```diff
@@ -86,7 +86,7 @@ check: manifests generate add-license lint test
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 .PHONY: test
 test: manifests generate fmt envtest check-license ## Run tests. Some test depend on Linux OS
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go run github.com/onsi/ginkgo/v2/ginkgo --label-filter="!integration" -coverprofile cover.out ./...
```
lukas016, line 89: we can install ginkgo as additional tool (similar way as controller-gen, kustomize, ...)

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,97 @@ func TestServer(t *testing.T) {
...
+	c := dialers.NewLocal()
+	libvirtConn = libvirt.NewWithDialer(c)
+	err = libvirtConn.Connect()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(libvirtConn.IsConnected(), BeTrue())
+
+	DeferCleanup(libvirtConn.ConnectClose)
 })
+
+func isSocketAvailable(socketPath string) (bool, error) {
```
lukas016, line 134: do you need boolean? You can return error, if socket is missing and otherwise it is nil

.github/workflows/integration-test.yml
afritzler: As we want to use public runners whenever possible we are bound to those one here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

provider/server/exec_test.go
```diff
@@ -28,8 +28,12 @@ var _ = Describe("Exec", func() {
 		Expect(resCreate).NotTo(BeNil())
 		machineID := resCreate.GetMachine().Metadata.Id
 
+		DeferCleanup(machineClient.DeleteMachine, &iri.DeleteMachineRequest{
```
lukas016, line 31: is this correct way how to use DeferCleanup function

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,97 @@ func TestServer(t *testing.T) {
...
+	DeferCleanup(cancel)
+
+	go func() {
+		defer GinkgoRecover()
+		Expect(app.Run(srvCtx, opts)).To(Succeed())
+	}()
+
+	Eventually(func() (bool, error) {
+		return isSocketAvailable(opts.Address)
+	}, "30s", "500ms").Should(BeTrue(), "The UNIX socket file should be available")
```
lukas016, line 114: suggestion: Eventually().WithTimeout(TIMEOUT).WithPolling(POLLING_INTERVAL)
it is more readable as old format for eventually which have generic interface: https://pkg.go.dev/github.com/onsi/gomega#Eventually

so-sahu:
go test result is inconsistent.

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,97 @@ func TestServer(t *testing.T) {
...
+	machineClient = iriv1alpha1.NewMachineRuntimeClient(gconn)
+	DeferCleanup(gconn.Close)
+
+	c := dialers.NewLocal()
+	libvirtConn = libvirt.NewWithDialer(c)
+	err = libvirtConn.Connect()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(libvirtConn.IsConnected(), BeTrue())
+
+	DeferCleanup(libvirtConn.ConnectClose)
```
so-sahu, line 131: I would suggest doing all the cleanup in the AfterSuite block, in place of using DeferCleanup in multiple places.

provider/server/machine_list_test.go
```diff
@@ -0,0 +1,48 @@
...
+		res, err := machineClient.CreateMachine(ctx, &iri.CreateMachineRequest{
+			Machine: &iri.Machine{
+				Metadata: &irimeta.ObjectMetadata{
+					Labels: map[string]string{
+						"machinepoolletv1alpha1.MachineUIDLabel": "foobar",
+					},
+				},
+				Spec: &iri.MachineSpec{
+					Power: iri.Power_POWER_ON,
+					// Image: &iri.ImageSpec{
```
so-sahu, line 25: Commented code can be removed.

provider/server/server_suite_test.go
afritzler: `DeferCleanup` has it's purpose if you e.g. are creating multiple `Machine`s and want to clean up ASAP since the GH runner instance has only limited capacity. Also there is no real need for an `AfterSuite` if you `DefereCleanup` where things are being defined: being it global clients in the suite setup or `It` block related resources.

so-sahu (changes requested):

provider/server/server_suite_test.go
so-sahu: ok, in that case DefereCleanup can be used for these purposes. However, I recommend having AfterSuite, where overall test artifacts cleanup has to be done.

provider/server/machine_create_test.go
```diff
@@ -24,13 +27,81 @@ var _ = Describe("CreateMachine", func() {
...
+		Expect(createResp).Should(SatisfyAll(
+			HaveField("Machine.Metadata.Id", Not(BeEmpty())),
+			HaveField("Machine.Spec.Power", iri.Power_POWER_ON),
+			HaveField("Machine.Spec.Image", Not(BeNil())),
+			HaveField("Machine.Spec.Class", machineClassx3xlarge),
+			HaveField("Machine.Spec.IgnitionData", Equal(ignitionData)),
+			HaveField("Machine.Spec.Volumes", BeNil()),
+			HaveField("Machine.Spec.NetworkInterfaces", BeNil()),
+			HaveField("Machine.Status.ObservedGeneration", BeZero()),
+			HaveField("Machine.Status.State", Equal(iri.MachineState_MACHINE_PENDING)),
```
so-sahu, line 52: Since the main purpose of this is to check if the machine object is created in the local store, I suggest not validating the status, if you are getting a PENDING state

provider/server/machine_create_test.go
```diff
@@ -24,13 +27,81 @@ var _ = Describe("CreateMachine", func() {
...
 			},
 		})
 		Expect(err).NotTo(HaveOccurred())
-		Expect(res).NotTo(BeNil())
+
+		By("ensuring the correct creation response")
+		Expect(createResp).Should(SatisfyAll(
+			HaveField("Machine.Metadata.Id", Not(BeEmpty())),
+			HaveField("Machine.Spec.Power", iri.Power_POWER_ON),
+			HaveField("Machine.Spec.Image", Not(BeNil())),
```
so-sahu, line 46: This can be equal to what ever image you have used in line 30.

pkg/controllers/machine_controller.go
```diff
@@ -431,6 +431,8 @@ func (r *MachineReconciler) domainFor(
 		return nil, nil, nil, err
 	}
 
+	domainSettings.Type = "kvm"
```
so-sahu, line 434: Could you please explain the context of explicitly setting this here? Also, I believe, this is not in the scope of this PR.

provider/server/server_suite_test.go
lukas016: Minimally i don't like how is defer DeferCleanup used. It completely ignoring errors returned from functions, here it isn't big problem but i think we have to check if machine deletion was successful.

pkg/controllers/machine_controller.go
kasabe28: Please look at my first comment for this.

so-sahu: The current code implementation is already setting the correct domain type. Below is a snippet from the domainXML generated by current controller FYR.
```
<domain type='kvm' id='163'>
```
Please explain what is the need of explicitly setting this here?

kasabe28: And what type of CPU is there with `kvm` domain type?

so-sahu:
```
<cpu mode='host-passthrough' check='none' migratable='on'/>
```

kasabe28: In our test environment we are getting domain type as `qemu` which is not supporting `host-passthrough` CPU. So If we want to support only `host-passthrough` CPU, we should limit domain type also to `kvm` or remove CPU limitation.

so-sahu: As I shared the snippets, the main logic is working fine.
I won't encourage to do changes in main logic to satisfy the test cases. I would suggest doing changes in test framework to get kvm as type, if that is what needed for test case.

lukas016: i agree with @so-sahu, we cannot change type of domain for test. It isn't compatible with our machines which currently exist. If you need different setup for tests, it has to be configurable and we need keep previous configuration as default.

Makefile
afritzler: We could. But we are installing tools into the `bin` folder for the sole reason to not having those dependencies in our `go.mod`. As `ginkgo` is need for testing and is hence a dependency this approach is valid.

provider/server/server_suite_test.go
afritzler: That is a valid point, we should use global constants for timeouts and interval.

provider/server/server_suite_test.go
afritzler: `DeferCleanup` does actually have proper error handling - so if the cleanup fails, your test will fail. Regarding the general usage of `DeferCleanup`: If you do the `Object` cleanup in an `AfterSuite` or `AfterEach` run, you need to keep track of when you created what which is a tedious task and you might forget something along the way leading to weird testing behaviours.

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,97 @@ func TestServer(t *testing.T) {
...
+	apinetPlugin.PluginName = "apinet"
+	apinetPlugin.AddFlags(fs)
+	Expect(fs.Set("apinet-node-name", "test-node")).NotTo(HaveOccurred())
+
+	opts := app.Options{
+		//TODO: set these using env variables if needed
+		Address:                     fmt.Sprintf("%s/test.sock", os.Getenv("PWD")),
+		BaseURL:                     baseURL,
+		PathSupportedMachineClasses: machineClassesFile.Name(),
+		RootDir:                     fmt.Sprintf("%s/libvirt-provider", userHomeDir),
```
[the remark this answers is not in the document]
afritzler: That is a valid point! We should use `GinkgoT().TempDir()` instead.

provider/server/machine_create_test.go
afritzler: Well the `Pending` state is the default/initial state of a `Machine`. I would say it is a valid point to check for this behaviour.

so-sahu: Agree. However, the machine status state is set in machine_controller.go.
Hence, in its test case it's good to validate the state. Moreover, we expect the machine to be in running state ideally when everything goes right. So, here it can be ignored. However, its optional. There is no harm in keeping. In that case. I would expect the machine to be in running state.

provider/server/server_suite_test.go
afritzler: The `ginkgo` community has also a take on this topic: https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup

> Setup and cleanup patterns like the one above are common in Ginkgo suites. While powerful, however, AfterEach nodes have a tendency to separate cleanup code from setup code. We had to create an originalWeightUnits closure variable to keep track of the original environment variable in the BeforeEach and pass it to the AfterEach - this feels noisy and potential error-prone.

Makefile
afritzler: That is correct we don't need CGO in this project as we are using `github.com/digitalocean/go-libvirt` a pure Go interface for interacting with libvirt.

provider/server/server_suite_test.go
so-sahu: Below can be the decision factors:
- **Resource Scope**: Consider whether cleanup is specific to individual tests or should be performed globally after all tests.
- **Isolation**: If tests need to be isolated and not share state, DeferCleanup might be more suitable.
- **Efficiency**: If cleanup can be consolidated and performed more efficiently once after all tests, AfterSuite might be preferred.

Makefile
lukas016: my main problem is name of variable. I don't know CGO variable and i expect it is typo for CGO_ENABLED=1|0

provider/server/server_suite_test.go
lukas016: @afritzler  DeferCleanup have proper handling of error but i don't see it in our code. You have to implement proper handlig of error with Expect() functions which are missing.

Makefile
afritzler: You are of course correct it should be `CGO_ENABLED` - we should either set it to 0 (to be on the safe side) or remove it as the default behaviour should be to build native without C support.

lukas016:
stop use amend and force push in active review process. THX

Makefile
lukas016: ok. i agree it isn't needed

provider/server/server_suite_test.go
lukas016: DeferCleanup are fixed

provider/server/machine_create_test.go
```diff
@@ -43,7 +43,7 @@ var _ = Describe("CreateMachine", func() {
 		Expect(createResp).Should(SatisfyAll(
 			HaveField("Machine.Metadata.Id", Not(BeEmpty())),
 			HaveField("Machine.Spec.Power", iri.Power_POWER_ON),
-			HaveField("Machine.Spec.Image", Not(BeNil())),
+			HaveField("Machine.Spec.Image.Image", Equal("ghcr.io/ironcore-dev/ironcore-image/gardenlinux:rootfs-dev-20231206-v1")),
```
so-sahu, line 46: Image name can be stored in a variable and used in multiple places.

provider/server/server_suite_test.go
```diff
@@ -4,31 +4,40 @@
...
 )
 
 const (
-	eventuallyTimeout    = 3 * time.Second
+	eventuallyTimeout    = 30 * time.Second
 	pollingInterval      = 50 * time.Millisecond
 	consistentlyDuration = 1 * time.Second
 	baseURL              = "http://localhost:8080"
 	machineClassx3xlarge = "x3-xlarge"
+	streamingAddress     = "127.0.0.1:20251"
```
lukas016, line 40: static port can be problematic on shared environment because multiple test cannot run in the same time. Are you okay with that?
@hardikdr can be baseURL different as streamingAddress? because you have note in flag:
```golang
fs.StringVar(&o.BaseURL, "base-url", "", "The base url to construct urls for streaming from. If empty it will be "+
		"constructed from the streaming-address")
```

it looks as dependency between baseURL and streamingAddress

provider/server/server_suite_test.go
sujeet01: Addressed.

Makefile
sujeet01: Done.

provider/server/machine_create_test.go
sujeet01: For better test flow readability, keeping it as is.

pkg/controllers/machine_controller.go
sujeet01: Removed the workaround, as it's no longer necessary. Apologies for the lengthy discussion and thanks for your patience!

provider/server/server_suite_test.go
sujeet01: Noted. Currently, the static port is not causing any issues, but we'll reassess if it causes any trouble with more tests.

provider/server/machine_create_test.go
so-sahu: It should be in the running state.

so-sahu (changes requested):
After completing the test execution, there are still two machines visible on the host. They need to be cleaned up.

so-sahu:
I have the following questions considering the current implementations:

- The testing relies on the host's libvirt. What happens if the test runs on a machine without libvirt installed?
- Is it advisable to utilize the host's libvirt for test case execution, considering the potential increase in its load?
- GinkgoT().TempDir() automatically cleans up temporary files and directories after each test. What if the user intends to retain these temporary files for inspection or debugging post-test execution?

provider/server/machine_create_test.go
kasabe28: In current `CreateMachine()` implementation we are not waiting for domain to be in `running` state.
https://github.com/ironcore-dev/libvirt-provider/blob/main/pkg/controllers/machine_controller.go#L268
As machine takes time to be in `running` state, we are checking for it later by listing machines.

so-sahu: In second reconciliation, its set to a running state. That is the overall design.

afritzler:
> * The testing relies on the host's libvirt. What happens if the test runs on a machine without libvirt installed?

That is why we have the `integration` test label when running those tests. So you can control when to run the integration test in an environment where `libvirt` is present (e.g. CI environment or a local box with libvirt installed). The default behaviour is to only run unit tests.

> * Is it advisable to utilize the host's libvirt for test case execution, considering the potential increase in its load?

I don't understand the problem here: you run those tests typically in throwaway test envs or your local box. Performance problems are not an issue here.

> * GinkgoT().TempDir() automatically cleans up temporary files and directories after each test. What if the user intends to retain these temporary files for inspection or debugging post-test execution?

In a local environment you could quickly change that to a non temp dir to retain the logs. In a GH environment where you typically can't access the runner file system you could use something like https://github.com/actions/upload-artifact to preserve logs or other artefacts of your job run for further investigation.

provider/server/machine_create_test.go
so-sahu: ok. Got your point. Make sense.

sujeet01:
> After completing the test execution, there are still two machines visible on the host. They need to be cleaned up.

Addressed.

hardikdr (commented):
Thanks for the PR. 

>After completing the test execution, there are still two machines visible on the host. They need to be cleaned up.

@so-sahu can you please confirm if this is resolved?

@sujeet01 @kasabe28 Have you been able to verify the framework with basic tests, such, 
- Creating Machine, verifying if the state turns Running for APIMachine. 
- Successful Machine creation with for a RAM boot, and etc.

provider/server/server_suite_test.go
```diff
@@ -38,38 +47,93 @@ func TestServer(t *testing.T) {
...
+	By("starting the app")
+
+	fs := pflag.NewFlagSet("test", pflag.ContinueOnError)
+	apinetPlugin := networkinterfaceplugin.NewDefaultOptions()
+	apinetPlugin.PluginName = "apinet"
+	apinetPlugin.AddFlags(fs)
+	Expect(fs.Set("apinet-node-name", "test-node")).To(Succeed())
+
+	tempDir := GinkgoT().TempDir()
+	Expect(os.Chmod(tempDir, 0777)).Should(Succeed())
```
lukas016, line 84: why 0777? it looks little dangerous and amateur.

so-sahu:
@hardikdr Yes, the cleanup is happening as expected now. However, I have seen some inconsistency, on which @sujeet01 is currently working.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install dependencies
+        run: |
+          set -x
+          sudo apt-get update
```
so-sahu, line 25: Optional: This can be made less verbose. eg: `sudo apt-get update -qq`

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,38 @@
...
+          sudo apt-get install -y --no-install-recommends qemu-utils ca-certificates libvirt-clients libcephfs-dev librbd-dev librados-dev libc-bin gcc
+          sudo apt-get install -y qemu-kvm libvirt-daemon-system
+          sudo systemctl enable --now libvirtd
+          sudo systemctl start libvirtd
+          sudo usermod -aG kvm $USER
+          sudo usermod -aG libvirt $USER
```
so-sahu, line 31: Optional: Both the usermod can be clubbed.

provider/server/machine_create_test.go
```diff
@@ -24,13 +28,86 @@ var _ = Describe("CreateMachine", func() {
...
+		By("ensuring domain for machine is in running state")
+		Eventually(func() libvirt.DomainState {
+			domainState, _, err := libvirtConn.DomainGetState(domain, 0)
+			Expect(err).NotTo(HaveOccurred())
+			return libvirt.DomainState(domainState)
+		}).Should(Equal(libvirt.DomainRunning))
+
+		By("updating machine annotations")
+		Eventually(func() error {
+			_, err := machineClient.UpdateMachineAnnotations(ctx, &iri.UpdateMachineAnnotationsRequest{
```
so-sahu, line 87: It is not needed in this test, as `UpdateMachineAnnotations` is not defined in machine_create.go. Rather, it can be properly validated in the test of machine_annotations_update.go. 
However, If you still wish to keep it here as well, please validate if the annotation is correctly set. But, that will be just a duplication of work.

provider/server/machine_create_test.go
```diff
@@ -24,13 +28,86 @@ var _ = Describe("CreateMachine", func() {
...
+			HaveField("Machine.Spec.Volumes", BeNil()),
+			HaveField("Machine.Spec.NetworkInterfaces", BeNil()),
+			HaveField("Machine.Status.ObservedGeneration", BeZero()),
+			HaveField("Machine.Status.State", Equal(iri.MachineState_MACHINE_PENDING)),
+			HaveField("Machine.Status.ImageRef", BeEmpty()),
+			HaveField("Machine.Status.Volumes", BeNil()),
+			HaveField("Machine.Status.NetworkInterfaces", BeNil()),
+		))
+
+		DeferCleanup(func(ctx SpecContext) {
```
so-sahu, line 59: Because the cleanup will be needed in multiple test files in this suite, I recommend moving common cleanup steps into a separate function for better readability and maintainability, which can be used everywhere else, to avoid rewriting code.

provider/server/machine_create_test.go
```diff
@@ -24,13 +28,86 @@ var _ = Describe("CreateMachine", func() {
...
+			domain, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
+			return err
+		}).Should(Succeed())
+		domainXMLData, err := libvirtConn.DomainGetXMLDesc(domain, 0)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(domainXMLData).NotTo(BeEmpty())
+
+		By("ensuring domain for machine is in running state")
+		Eventually(func() libvirt.DomainState {
+			domainState, _, err := libvirtConn.DomainGetState(domain, 0)
```
so-sahu, line 80: Similarly, the domain state check can be extracted into a separate function for code reuse.

provider/server/machine_create_test.go
```diff
@@ -4,17 +4,21 @@
...
 
 var _ = Describe("CreateMachine", func() {
 
-	It("should correctly create a machine", func(ctx SpecContext) {
+	It("should create a machine", func(ctx SpecContext) {
 		By("creating a machine")
-		res, err := srv.CreateMachine(ctx, &iri.CreateMachineRequest{
+		ignitionData := []byte("urjhikmnbdjfkknhhdddeee")
+		imageURL := "ghcr.io/ironcore-dev/ironcore-image/gardenlinux:rootfs-dev-20231206-v1"
+		createResp, err := machineClient.CreateMachine(ctx, &iri.CreateMachineRequest{
```
so-sahu, line 21: Similarly, the machine creation can be extracted into a separate function for code reuse.

lukas016:
guys did you think about different concept of integration tests? Currently integration tests and unit tests are written the same way and label is only difference, so reading of tests is more harder because names of files are same (suffix _test.go). We had little different concept base on blackbox testing. Where we deploy/start application independent on test in target env and our integration test simulate client behavior. Main benefits is separate integration test into separate folder (e2e) and testing whole application from start to end (parsing arguments, graceful shutdown, ...). Of course it has disadvantages mainly you cannot analyze logs easy in integration test and it is really blackbox test.

provider/server/machine_create_test.go
kasabe28: Currently CreateMachine() always returns `PENDING` state for new machine. For more details please refer this: https://github.com/ironcore-dev/libvirt-provider/issues/102
Once this gets fixed we wont need `UpdateMachineAnnotations` anymore here. As of now its been added as an workaround to get latest domain state and update it in machinestore.

provider/server/server_suite_test.go
sujeet01: In the actual codebase, where the user's home directory serves as the root directory, special permissions are not required. However, during testing, we create a temporary directory which typically gets created at root's /tmp, requiring broader permissions for successful execution. I've optimized the tests to operate with the minimal required permissions of 0703.

provider/server/machine_create_test.go
so-sahu: I get the machine running state. So, I would like to differ here. As per the design, 1st reconciliation sets the machine state to pending and it gets the running state in the next iteration. 
The controller is working fine as I have observed. 

Let's take this offline.

provider/server/server_suite_test.go
lukas016: but still why everyone has permission for write and execute, it is technically more dangerous part of permission.

provider/server/machine_create_test.go
so-sahu: @afritzler Could you please let us know your view on this.

lukas016: it is more nice to have for me but it isn't important. We can do improvement always later and i prefer working tests.

provider/server/server_suite_test.go
sujeet01: Addressed, removed others permissions.

provider/server/exec_test.go
```diff
@@ -25,11 +27,36 @@ var _ = Describe("Exec", func() {
...
 		Expect(err).NotTo(HaveOccurred())
-		Expect(resCreate).NotTo(BeNil())
-		machineID := resCreate.GetMachine().Metadata.Id
+		Expect(createResp).NotTo(BeNil())
+
+		DeferCleanup(func(ctx SpecContext) {
+			_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+			Expect(err).ShouldNot(HaveOccurred())
+			Eventually(func() bool {
+				_, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
```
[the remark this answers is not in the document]
sujeet01: Choosing to use `createResp.Machine.Metadata.Id` directly throughout the code is intentional, as it is more descriptive and clear, improving code readability. Introducing a separate variable like `machineID` would add unnecessary indirection.

provider/server/machine_create_test.go
sujeet01: The direct use of `machineClient.CreateMachine` in the test ensures clear and straightforward readability. Adding another function may introduce unnecessary abstraction without significant improvements in clarity.

provider/server/exec_test.go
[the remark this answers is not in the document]
so-sahu: Introducing a named variable could contribute to code clarity and simplicity. However, as its not using any expression or repeated calls to any functions, it completely depends on one's coding practice. Hence, Its completely optional in your case.

provider/server/machine_create_test.go
so-sahu: I suggest reusing the code instead of duplicating it across multiple places. Providing a meaningful name will enhance code clarity. While utilizing machineClient.CreateMachine directly aligns with your reasons, it is advisable to centralize the machine definition in one location for broader usage.

provider/server/server_suite_test.go
lukas016: how is it removed, when you returned worse version with 0777? Do you know how permissions on linux work?

Makefile
```diff
@@ -175,6 +175,10 @@ addlicense: $(ADDLICENSE) ## Download addlicense locally if necessary.
 $(ADDLICENSE): $(LOCALBIN)
 	test -s $(LOCALBIN)/addlicense || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@$(ADDLICENSE_VERSION)
 
+.PHONY: integration-tests
```
lukas016, line 178: move this target near standard unit test target. This section primary for installation additional tools

lukas016: pls add some comment for use make help

provider/server/server_suite_test.go
lukas016: sorry, it was fixed. github opened to me wrong version.

kasabe28:
> Thanks for the PR.
> 
> > After completing the test execution, there are still two machines visible on the host. They need to be cleaned up.
> 
> @so-sahu can you please confirm if this is resolved?
> 
> @sujeet01 @kasabe28 Have you been able to verify the framework with basic tests, such,
> 
> * Creating Machine, verifying if the state turns Running for APIMachine.
> * Successful Machine creation with for a RAM boot, and etc.

Yes, current framework supports creating machine with RAM boot and we are also able to verify `APIMachine` state turns to `Running`.

kasabe28:
> guys did you think about different concept of integration tests? Currently integration tests and unit tests are written the same way and label is only difference, so reading of tests is more harder because names of files are same (suffix _test.go). We had little different concept base on blackbox testing. Where we deploy/start application independent on test in target env and our integration test simulate client behavior. Main benefits is separate integration test into separate folder (e2e) and testing whole application from start to end (parsing arguments, graceful shutdown, ...). Of course it has disadvantages mainly you cannot analyze logs easy in integration test and it is really blackbox test.

As discussed offline current solution meets the requirements of integration tests

Makefile
sujeet01: Done.

provider/server/machine_create_test.go
kasabe28: Removed `UpdateMachineAnnotations` workaround and added empty disk volume to create machine successfully. We can have test without any volume when #102 gets fixed.

hardikdr (approved):
@so-sahu @lukas016 can you please provide a final review?

so-sahu (approved):
LGTM

lukas016 (approved):

PR #101, opened 2024-01-05
Implementing Graceful Shutdown for VMs

### LINKED ISSUE #73: Implement Graceful Shutdown for VMs
# Summary

Currently, libvirt-provider destroys VMs directly, potentially causing data loss or corruption as applications running inside the VM may not have adequate time for clean-up and a graceful shutdown.

It would be highly beneficial if libvirt-provider could support a graceful shutdown process for VMs. This process should involve sending an ACPI power event to the guest operating system within the VM. If the guest OS supports ACPI events, applications and processes inside the VM should close gracefully.

## Basic example

libvirt-provider could leverage libvirt's `virsh shutdown` feature to initiate the graceful shutdown and follow up with a `virsh destroy` command after a timeout period if the VM hasn't shut down.

Need to confirm if Gardenlinux supports ACPI and has the necessary ACPI handlers/daemons to process such a shutdown request.

## Motivation

Enhance data safety and integrity for applications running in VMs.

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/73

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -10,9 +10,10 @@ type Metadata struct {
 	Annotations map[string]string `json:"annotations"`
 	Labels      map[string]string `json:"labels"`
 
-	CreatedAt  time.Time  `json:"createdAt"`
-	DeletedAt  *time.Time `json:"deletedAt,omitempty"`
-	Generation int64      `json:"generation"`
+	CreatedAt       time.Time  `json:"createdAt"`
+	DeletedAt       *time.Time `json:"deletedAt,omitempty"`
+	FirstShutdownAt time.Time  `json:"firstShutdownAt"`
+	Generation      int64      `json:"generation"`
 
 	Finalizers []string `json:"finalizers,omitempty"`
 }
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -10,10 +10,9 @@ type Metadata struct {
 	Annotations map[string]string `json:"annotations"`
 	Labels      map[string]string `json:"labels"`
 
-	CreatedAt       time.Time  `json:"createdAt"`
-	DeletedAt       *time.Time `json:"deletedAt,omitempty"`
-	FirstShutdownAt time.Time  `json:"firstShutdownAt"`
-	Generation      int64      `json:"generation"`
+	CreatedAt  time.Time  `json:"createdAt"`
+	DeletedAt  *time.Time `json:"deletedAt,omitempty"`
+	Generation int64      `json:"generation"`
 
 	Finalizers []string `json:"finalizers,omitempty"`
 }
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -42,7 +42,6 @@ const (
 	rootFSAlias                     = "ua-rootfs"
 	libvirtDomainXMLIgnitionKeyName = "opt/com.coreos/config"
 	networkInterfaceAliasPrefix     = "ua-networkinterface-"
-	defaultRetryReconcileTime       = 15 * time.Second
 )
 
 var (
@@ -316,7 +315,7 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
 			var destroid bool
 
 			if !machine.FirstShutdownAt.IsZero() && time.Now().After(machine.FirstShutdownAt.Add(r.vmGracefulShutdownTimeout)) {
-				destroid, err = r.destroyAndCleanupMachine(ctx, logger, machine)
+				destroid, err = r.destroyMachine(ctx, logger, machine)
 			} else {
 				err = r.processShutdown(ctx, logger, machine)
 			}
@@ -336,19 +335,17 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
 	}, r.resyncIntervalGarbageCollector)
 }
 
-func (r *MachineReconciler) destroyAndCleanupMachine(ctx context.Context, log logr.Logger, machine *api.Machine) (bool, error) {
-	log.V(1).Info("Starting deletion")
-
+func (r *MachineReconciler) destroyMachine(ctx context.Context, log logr.Logger, machine *api.Machine) (bool, error) {
 	err := r.destroyDomain(log, machine)
 	if err != nil {
-		return false, fmt.Errorf("failed to destory machine domain: %w", err)
+		return false, fmt.Errorf("failed to destroy machine domain: %w", err)
 	}
 	return true, nil
 
 }
 
 func (r *MachineReconciler) processShutdown(ctx context.Context, log logr.Logger, machine *api.Machine) error {
-	log.V(1).Info("Shutting Down Machine")
+	log.V(1).Info("Shutting down Machine")
 
 	if machine.FirstShutdownAt.IsZero() {
 		machine.FirstShutdownAt = time.Now()
@@ -370,6 +367,8 @@ func (r *MachineReconciler) processShutdown(ctx context.Context, log logr.Logger
 }
 
 func (r *MachineReconciler) destroyDomain(log logr.Logger, machine *api.Machine) error {
+	log.V(1).Info("Starting domain destroy")
+
 	domain := libvirt.Domain{
 		UUID: libvirtutils.UUIDStringToBytes(machine.ID),
 	}
@@ -383,30 +382,30 @@ func (r *MachineReconciler) destroyDomain(log logr.Logger, machine *api.Machine)
 }
 
 func (r *MachineReconciler) cleanupAdditionalResources(ctx context.Context, log logr.Logger, machine *api.Machine) error {
+	log.V(1).Info("Starting additional resources cleanup")
+
 	if err := r.deleteVolumes(ctx, log, machine); err != nil {
 		return fmt.Errorf("failed to remove machine disks: %w", err)
 	}
-
 	log.V(1).Info("Removed machine disks")
 
 	if err := r.deleteNetworkInterfaces(ctx, log, machine); err != nil {
 		return fmt.Errorf("failed to remove machine network interfaces: %w", err)
 	}
-
 	log.V(1).Info("Removed network interfaces")
 
 	if err := os.RemoveAll(r.host.MachineDir(machine.ID)); err != nil {
 		return fmt.Errorf("failed to remove machine directory: %w", err)
 	}
-
 	log.V(1).Info("Removed machine directory")
 
 	machine.Finalizers = utils.DeleteSliceElement(machine.Finalizers, MachineFinalizer)
 	if _, err := r.machines.Update(ctx, machine); store.IgnoreErrNotFound(err) != nil {
 		return fmt.Errorf("failed to update machine metadata: %w", err)
 	}
-
 	log.V(1).Info("Removed Finalizers")
+
+	log.V(1).Info("Removed additional resources")
 	log.V(1).Info("Deletion completed")
 	return nil
 }
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -4,6 +4,7 @@
 package controllers
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"encoding/xml"
@@ -25,7 +26,7 @@ import (
 	"github.com/ironcore-dev/libvirt-provider/pkg/os/osutils"
 	providernetworkinterface "github.com/ironcore-dev/libvirt-provider/pkg/plugins/networkinterface"
 	providervolume "github.com/ironcore-dev/libvirt-provider/pkg/plugins/volume"
-	providerhost "github.com/ironcore-dev/libvirt-provider/pkg/providerhost" // TODO: Change to a better naming for all imports, libvirthost?
+	providerhost "github.com/ironcore-dev/libvirt-provider/pkg/providerhost"
 	"github.com/ironcore-dev/libvirt-provider/pkg/raw"
 	"github.com/ironcore-dev/libvirt-provider/pkg/store"
 	"github.com/ironcore-dev/libvirt-provider/pkg/utils"
@@ -42,6 +43,7 @@ const (
 	rootFSAlias                     = "ua-rootfs"
 	libvirtDomainXMLIgnitionKeyName = "opt/com.coreos/config"
 	networkInterfaceAliasPrefix     = "ua-networkinterface-"
+	defaultRetryReconcileTime       = 15 * time.Second
 )
 
 var (
@@ -56,6 +58,8 @@ var (
 		//libvirt.DomainShutoff:     api.MachineStateShutdown,
 		libvirt.DomainPmsuspended: api.MachineStatePending,
 	}
+
+	errNeedToRequeue = errors.New("need to requeue")
 )
 
 type MachineReconcilerOptions struct {
@@ -70,6 +74,7 @@ type MachineReconcilerOptions struct {
 	ResyncIntervalVolumeSize   time.Duration
 	ResyncIntervalMachineState time.Duration
 	EnableHugepages            bool
+	VMGracefulShutdownTimeout  time.Duration
 }
 
 func setMachineReconcilerOptionsDefaults(o *MachineReconcilerOptions) {
@@ -115,6 +120,7 @@ func NewMachineReconciler(
 		resyncIntervalVolumeSize:   opts.ResyncIntervalVolumeSize,
 		resyncIntervalMachineState: opts.ResyncIntervalMachineState,
 		enableHugepages:            opts.EnableHugepages,
+		vmGracefulShutdownTimeout:  opts.VMGracefulShutdownTimeout,
 	}, nil
 }
 
@@ -138,6 +144,8 @@ type MachineReconciler struct {
 
 	resyncIntervalVolumeSize   time.Duration
 	resyncIntervalMachineState time.Duration
+
+	vmGracefulShutdownTimeout time.Duration
 }
 
 func (r *MachineReconciler) Start(ctx context.Context) error {
@@ -323,7 +331,6 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 		if err := r.deleteMachine(ctx, log, machine); err != nil {
 			return fmt.Errorf("failed to delete machine: %w", err)
 		}
-		log.V(1).Info("Successfully deleted machine")
 		return nil
 	}
 
@@ -402,6 +409,116 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger, 
 
 	log.V(1).Info("Finalizer present, doing cleanup")
 
+	domain, err := r.libvirt.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(machine.ID))
+	if err != nil {
+
+		if libvirt.IsNotFound(err) {
+			err = r.cleanupAdditionalResources(ctx, log, machine)
+			if err != nil {
+				r.queue.AddAfter(machine.ID, defaultRetryReconcileTime)
+				return err
+			}
+			log.V(1).Info("Cleaned up all dependencies, removing finalizer")
+			machine.Finalizers = utils.DeleteSliceElement(machine.Finalizers, MachineFinalizer)
+			if _, err := r.machines.Update(ctx, machine); store.IgnoreErrNotFound(err) != nil {
+				return fmt.Errorf("failed to update machine metadata: %w", err)
+			}
+			log.V(1).Info("Successfully removed finalizer")
+			log.V(1).Info("Successfully deleted machine")
+			return nil
+		}
+		return fmt.Errorf("cannot get machine domain: %w", err)
+	}
+
+	err = r.shutdownDomain(log, domain, machine.ID)
+	if err != nil {
+		if errors.Is(err, errNeedToRequeue) {
+			log.V(1).Info("Requeue: shutdown graceful period")
+			err = nil
+		}
+		r.queue.AddAfter(machine.ID, defaultRetryReconcileTime)
+		return err
+	}
+
+	// issue: this can be affect performance, read function comments
+	err = r.destroyDomain(log, domain)
+	if err != nil {
+		err = fmt.Errorf("destruction of Machine domain failed: %w", err)
+	}
+
+	r.queue.AddAfter(machine.ID, defaultRetryReconcileTime)
+	return err
+}
+
+func (r *MachineReconciler) shutdownDomain(log logr.Logger, domain libvirt.Domain, machineUID string) error {
+	domainXML, err := r.getDomainDesc(machineUID)
+	if err != nil {
+		return fmt.Errorf("error getting machine description: %w", err)
+	}
+
+	providerMeta := &providermeta.LibvirtProviderMetadata{}
+	err = xml.Unmarshal([]byte(domainXML.Metadata.XML), providerMeta)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling libvirt-provider metadata: %w", err)
+	}
+
+	if providerMeta.ShutdownTimestamp.IsZero() {
+		log.V(1).Info("Shutdown machine")
+		// If the shutdown command fails persistently, it could lead to repeated re-queueing without resolution as domain metadata is not patched.
+
+		err = r.libvirt.DomainShutdownFlags(domain, libvirt.DomainShutdownAcpiPowerBtn)
+		if err != nil {
+			return fmt.Errorf("domain shutdown failed: %w", err)
+		}
+
+		providerMeta.ShutdownTimestamp = time.Now()
+		err := r.updateDomainMetadata(log, domain, providerMeta)
+		if err != nil {
+			return fmt.Errorf("update of domain metadata failed: %w", err)
+		}
+
+		return errNeedToRequeue
+	}
+
+	if providerMeta.ShutdownTimestamp.Add(r.vmGracefulShutdownTimeout).After(time.Now()) {
+		return errNeedToRequeue
+	}
+
+	return nil
+}
+
+// destroyDomain will do graceful destroy of domain
+func (r *MachineReconciler) destroyDomain(log logr.Logger, domain libvirt.Domain) error {
+	log.V(1).Info("Destroying domain")
+	// DomainDestroyFlags is blocking, so it can create performance issue in future
+	// During test with 26 emptyDisks function call spent max 1s.
+	// Performance issue will be in edge cases, if machine is correctly booted and it ignored shutdown command.
+	err := r.libvirt.DomainDestroyFlags(domain, libvirt.DomainDestroyGraceful)
+	if err != nil && !libvirt.IsNotFound(err) {
+		return err
+	}
+
+	log.V(1).Info("Successfully destroyed domain")
+	return nil
+}
+
+func (r *MachineReconciler) updateDomainMetadata(log logr.Logger, domain libvirt.Domain, metadata *providermeta.LibvirtProviderMetadata) error {
+	log.V(1).Info("Updating domain metadata")
+	// Only 3 fields have flexible size.
+	// Range is between 245 - 291 bytes
+	// To prevent buffer reallocation, extra bytes are included in the initial size.
+	const xmlLength = 350
+	buf := bytes.NewBuffer(make([]byte, 0, xmlLength))
+	enc := xml.NewEncoder(buf)
+	err := metadata.MarshalXMLWithoutNamespace(enc)
+	if err != nil {
+		return err
+	}
+
+	return r.libvirt.DomainSetMetadata(domain, int32(libvirt.DomainMetadataElement), libvirt.OptString{buf.String()}, libvirt.OptString{"libvirtprovider"}, libvirt.OptString{"https://github.com/ironcore-dev/libvirt-provider"}, libvirt.DomainAffectCurrent)
+}
+
+func (r *MachineReconciler) cleanupAdditionalResources(ctx context.Context, log logr.Logger, machine *api.Machine) error {
 	log.V(1).Info("Removing volumes")
 	if err := r.deleteVolumes(ctx, log, machine); err != nil {
 		return fmt.Errorf("error removing machine disks: %w", err)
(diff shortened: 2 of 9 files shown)


pkg/controllers/machine_controller.go
```diff
@@ -290,65 +298,133 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger,
...
+		if libvirt.IsNotFound(err) {
+			err = r.cleanupAdditionalResources(ctx, log, machine)
+			if err != nil {
+				r.queue.AddAfter(machine.ID, defaultRetryReconcileTime)
+				return err
+			}
+			log.V(1).Info("Cleaned up all dependencies, removing finalizer")
+			machine.Finalizers = utils.DeleteSliceElement(machine.Finalizers, MachineFinalizer)
+			if _, err := r.machines.Update(ctx, machine); store.IgnoreErrNotFound(err) != nil {
+				return fmt.Errorf("failed to update machine metadata: %w", err)
```
lukas016, line 313: will it requeue again?

so-sahu: yes, the control will go to line 198, and item will be requeued.
(https://github.com/ironcore-dev/libvirt-provider/blob/8c953a3bdf012e2ff2a4f6732027a21679b4e7cd/pkg/controllers/machine_controller.go#L198C3-L198C31)

pkg/controllers/machine_controller.go
```diff
@@ -290,65 +298,133 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger,
...
 
-func (r *MachineReconciler) deleteDomain(log logr.Logger, machine *api.Machine) error {
-	domain := libvirt.Domain{
-		UUID: libvirtutils.UUIDStringToBytes(machine.ID),
+// destroyDomain will do graceful destroy of domain
+func (r *MachineReconciler) destroyDomain(log logr.Logger, domain libvirt.Domain) error {
+	log.V(1).Info("Destroying domain")
+	// DomainDestroyFlags is blocking, so it can create performance issue in future
+	// During test with 26 emptyDisks function call spent max 1s.
+	// Performance issue will be in edge cases, if machine is correctly booted and it ignored shutdown command.
```
lukas016, line 384:
```suggestion
	// Performance issue will be in edge cases, if machine ignore shutdown command (not booted,  ignoring acpi signals, ...).
```

so-sahu: ok. I will change it.

pkg/meta/meta.go
```diff
@@ -7,10 +7,12 @@ import (
 	"encoding/xml"
 	"fmt"
 	"strings"
+	"time"
 )
 
 type LibvirtProviderMetadata struct {
-	IRIMmachineLabels string `xml:"irimachinelabels"`
+	IRIMmachineLabels string    `xml:"irimachinelabels"`
```
lukas016, line 14: it looks as mixing of 2 key format without _ and with _. I prefer unified naming convetion

lukas016:
@lukasfrank if i remember correctly, you implemented exponential requeue for every returned error. Is it correct? Because if it is true, we can simplify code and delete requeue error and custom requeue interval.

lukasfrank:
> @lukasfrank if i remember correctly, you implemented exponential requeue for every returned error. Is it correct? Because if it is true, we can simplify code and delete requeue error and custom requeue interval.

Correct - the requeuing is already implemented in the `RateLimitingQueue` ( `k8s.io/utils/clock`)

so-sahu:
I acknowledge that the requeuing functionality is already incorporated in the RateLimitingQueue. However, I find the nanosecond precision too rapid, especially in the context of virtual machine shutdowns (refer to: [source](https://github.com/kubernetes/client-go/blob/fb8b7346aacefea5ee2ab2e234afc4451c90c435/util/workqueue/default_rate_limiters.go#L97)). When utilizing the RateLimitingQueue for VM shutdowns, the heightened precision leads to accelerated reconciliations, resulting in an unnecessary surge in the number of machine reconciliations.

Therefore, I propose the inclusion of existing custom requeue interval for the graceful shutdown of VMs.

lukasfrank:
> I acknowledge that the requeuing functionality is already incorporated in the RateLimitingQueue. However, I find the nanosecond precision too rapid, especially in the context of virtual machine shutdowns (refer to: [source](https://github.com/kubernetes/client-go/blob/fb8b7346aacefea5ee2ab2e234afc4451c90c435/util/workqueue/default_rate_limiters.go#L97)). When utilizing the RateLimitingQueue for VM shutdowns, the heightened precision leads to accelerated reconciliations, resulting in an unnecessary surge in the number of machine reconciliations.
> 
> Therefore, I propose the inclusion of existing custom requeue interval for the graceful shutdown of VMs.

@so-sahu I think you mixed something up: The delay is only converted to nanoseconds. We use the following (default) settings: 

```
func DefaultControllerRateLimiter() RateLimiter {
	return NewMaxOfRateLimiter(
		NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
		// 10 qps, 100 bucket size.  This is only for retry speed and its only the overall factor (not per item)
		&BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
	)
}
```

pkg/controllers/machine_controller.go
```diff
@@ -40,6 +42,7 @@ const (
 	rootFSAlias                     = "ua-rootfs"
 	libvirtDomainXMLIgnitionKeyName = "opt/com.coreos/config"
 	networkInterfaceAliasPrefix     = "ua-networkinterface-"
+	defaultRetryReconcileTime       = 15 * time.Second
```
lukas016, line 45: Do we need this reconciletime?

lukas016:
```text
5
10
20
40
80
160
320
640
1280 (1.28s)
2560
5120
10240
20480
40960
81920
163840
327680
655360
```
@lukasfrank as you can see, we currently start with 5ms timeout between reconcile loops. So if we will return error for reconcilation, it will call another reconcile loops under 1seconds what is quick for shutdown process because machine need minimal seconds for graceful shutdown. So sourav has problem with this first 10 reconciliation  loops.

lukas016:
> > I acknowledge that the requeuing functionality is already incorporated in the RateLimitingQueue. However, I find the nanosecond precision too rapid, especially in the context of virtual machine shutdowns (refer to: [source](https://github.com/kubernetes/client-go/blob/fb8b7346aacefea5ee2ab2e234afc4451c90c435/util/workqueue/default_rate_limiters.go#L97)). When utilizing the RateLimitingQueue for VM shutdowns, the heightened precision leads to accelerated reconciliations, resulting in an unnecessary surge in the number of machine reconciliations.
> > Therefore, I propose the inclusion of existing custom requeue interval for the graceful shutdown of VMs.
> 
> @so-sahu I think you mixed something up: The delay is only converted to nanoseconds. We use the following (default) settings:
> 
> ```
> func DefaultControllerRateLimiter() RateLimiter {
> 	return NewMaxOfRateLimiter(
> 		NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
> 		// 10 qps, 100 bucket size.  This is only for retry speed and its only the overall factor (not per item)
> 		&BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
> 	)
> }
> ```

Hi, @lukasfrank and @so-sahu i did mistake when i asked this as comment. i created conversation for continue discussion about this problem (https://github.com/ironcore-dev/libvirt-provider/pull/101#discussion_r1448436933) because using standard comments impact readability of PR bad way.

lukasfrank:
As part of an offline discussion (@so-sahu  @lukas016)  we can conclude that the current approach is misusing the reconcile loop to delay the deletion process which leads to odd and unstable 

The agreed second iteration behaves the following way: 
- [x]  Server adds the deletion timestamp (already there) 
- [ ]  Additional routine checks periodically ( `store.List`) if machines in the store contain deletion timestamp 
  - [ ] If timestamp + shutdown period is exceed the machine is being shutdown forcefully 
  - [ ] Otherwise VM receives graceful shutdown signal 
- [ ] machine dependent resources are being deleted + remove finalizer

To summarise: The shutdown handling is being factored out of the main routine + the main routine stays almost the same

lukasfrank (changes requested):
As part of an offline discussion (@so-sahu  @lukas016)  we can conclude that the current approach is misusing the reconcile loop to delay the deletion process which leads to odd and unstable behaviour. 

The agreed second iteration behaves the following way: 
- [x]  Server adds the deletion timestamp (already there) 
- [ ]  Additional routine checks periodically ( `store.List`) if machines in the store contain deletion timestamp 
  - [ ] If timestamp + shutdown period is exceed the machine is being shutdown forcefully 
  - [ ] Otherwise VM receives graceful shutdown signal 
- [ ] machine dependent resources are being deleted + remove finalizer

To summarise: The shutdown handling is being factored out of the main routine + the main routine stays almost the same

provider/cmd/app/app.go
```diff
@@ -117,6 +119,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.Libvirt.Qcow2Type, "qcow2-type", qcow2.Default(), fmt.Sprintf("qcow2 implementation to use. Available: %v", qcow2.Available()))
 
+	fs.DurationVar(&o.VMGracefulShutdownTimeout, "vm-graceful-shutdown-timeout", 5*time.Minute, "Sets graceful time period for shutdown of VM. If the machine isn't shutdown in time, vm will be deleted with force process.")
```
afritzler, line 122:
```suggestion
	fs.DurationVar(&o.VMGracefulShutdownTimeout, "vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly terminated.")
```

afritzler (changes requested):
Instead of requeuing the `Machine` when we shutdown the domain I would rather implement the following termination flow:
- When we delete a `Machine`, we add the `deletionTimestamp` to the `Machine`s metadata and initiate a domain shutdown
- In parallel we start in the `app.go` a GC go routine (e.g. a Garbage collection process) which configurable via flag (e.g. `--gc-resync-period`) is running over all `Machines` in the local store and checks if `deletionTimestamt + GracePeriod` > `currentTime`. If that is the case we forcefully shutdown the domain.

I would not reuse the requeuing mechanism in the reconciler for doing cleanup jobs.

UPDATE: sorry I didn't see @lukasfrank comment. But essentially this goes into the same direction.

pkg/controllers/machine_controller.go
```diff
@@ -290,65 +298,133 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger,
...
+	if err := r.deleteNetworkInterfaces(ctx, log, machine); err != nil {
+		return fmt.Errorf("error removing machine network interfaces: %w", err)
+	}
+	log.V(1).Info("Successfully removed network interfaces")
+
+	log.V(1).Info("Removing machine directory")
+	if err := os.RemoveAll(r.host.MachineDir(machine.ID)); err != nil {
+		return fmt.Errorf("error removing machine directory: %w", err)
+	}
+	log.V(1).Info("Successfully removed machine directory")
```
afritzler, line 427: We should avoid logging which printing "successful" in the message. Instead we should use e.g. "removing XYZ" when initiating an action and "removed XYZ" when the process is finished. The definition of what "successful" means is most of the time ambiguous.

afritzler (changes requested):

provider/cmd/app/app.go
```diff
@@ -117,6 +120,9 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.Libvirt.Qcow2Type, "qcow2-type", qcow2.Default(), fmt.Sprintf("qcow2 implementation to use. Available: %v", qcow2.Available()))
 
+	fs.DurationVar(&o.VMGracefulShutdownTimeout, "vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly terminated.")
```
lukas016, line 123: what do you think about gc-vm-graceful-shutdown-period/timeout. Mainly i want to know gc-resync-interval and this argument have some relationship.

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +189,118 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 	return nil
 }
 
+func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.Logger) {
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
+		machines, err := r.machines.List(ctx)
```
lukas016, line 194: i think we need recovery logic for panic. this function has to be design as failure tolerant

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +189,118 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+
+			if err != nil {
+				logger.Error(err, "failed to garbage collector process")
+			}
+		}
+
+	}, r.resyncIntervalGarbageCollector)
+}
+
+func (r *MachineReconciler) destroyAndCleanupMachine(ctx context.Context, log logr.Logger, machine *api.Machine) (bool, error) {
```
lukas016, line 229: rename function. cleanup isn't anymore part of this function

lukas016 (changes requested):

pkg/controllers/machine_controller.go
lukas016: recovery logic is part of wait.UntilWithContext. it little looks behavior of panic propagation to top is manage be variable ReallyCrash which is set to true by default: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/runtime/runtime.go#L34

provider/cmd/app/app.go
so-sahu: Changed

pkg/controllers/machine_controller.go
so-sahu: Thanks for your suggestion. It is now followed.

provider/cmd/app/app.go
so-sahu: I would prefer to mention the link in the description. Something like: `Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly destroyed by garbage collector`

pkg/controllers/machine_controller.go
lukas016: minimally panic has to be logged in same format as other log records.

garbage collector is for me core functionality of libvirt-provider which has to work. But i prefer here really failure tolerant logic. Main reason is affect another more important logic (create machine). When machine cannot be deleted okay it is problem but you can still create new machines and you have time solve root cause with GC. If panic crash application, app can be stuck in infinity restart loop and i think it is worse case.

so-sahu: As discussed privately, there is no need for us to be concerned about this at the moment.

provider/cmd/app/app.go
so-sahu: Added it to the description, as well as the name is changed.

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
so-sahu: removed.

pkg/controllers/machine_controller.go
so-sahu: done.

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +189,118 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+	log.V(1).Info("Starting deletion")
```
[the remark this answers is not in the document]
so-sahu: done

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +189,118 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+	err := r.destroyDomain(log, machine)
+	if err != nil {
+		return false, fmt.Errorf("failed to destory machine domain: %w", err)
+	}
+	return true, nil
+
+}
+
+func (r *MachineReconciler) processShutdown(ctx context.Context, log logr.Logger, machine *api.Machine) error {
+	log.V(1).Info("Shutting Down Machine")
```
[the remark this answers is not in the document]
so-sahu: changed

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +189,118 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+	if err := r.libvirt.DomainDestroy(domain); err != nil && !libvirtutils.IsErrorCode(err, libvirt.ErrNoDomain) {
+		return fmt.Errorf("failed to destroy domain: %w", err)
+	}
+
+	log.V(1).Info("Destroyed domain")
+	return nil
+}
+
+func (r *MachineReconciler) cleanupAdditionalResources(ctx context.Context, log logr.Logger, machine *api.Machine) error {
+	if err := r.deleteVolumes(ctx, log, machine); err != nil {
```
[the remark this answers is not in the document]
so-sahu: followed.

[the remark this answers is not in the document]
lukas016: maybe add starting log before every deletion step (volume, nic, machinedir) can be useful for ops guys too

provider/cmd/app/app.go
```diff
@@ -120,7 +120,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.Libvirt.Qcow2Type, "qcow2-type", qcow2.Default(), fmt.Sprintf("qcow2 implementation to use. Available: %v", qcow2.Available()))
 
-	fs.DurationVar(&o.VMGracefulShutdownTimeout, "vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly terminated.")
+	fs.DurationVar(&o.VMGracefulShutdownTimeout, "gc-vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly destroyed by garbage collector.")
 	fs.DurationVar(&o.ResyncIntervalGarbageCollector, "gc-resync-interval", 1*time.Minute, "Duration for garbage collector resynchronization.")
```
lukas016, line 124: I suggest here replace duration with interval. Duration could be explained as timeout for GC process.

provider/cmd/app/app.go
```diff
@@ -120,7 +120,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.Libvirt.Qcow2Type, "qcow2-type", qcow2.Default(), fmt.Sprintf("qcow2 implementation to use. Available: %v", qcow2.Available()))
 
-	fs.DurationVar(&o.VMGracefulShutdownTimeout, "vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly terminated.")
```
lukas016: optional: renaming of variables to same design as flags can be useful for programmers.

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
so-sahu: I sense an excessive amount of logging. Nevertheless, the absence of the completed log signifies an issue. Fortunately, we now possess the log indicating the initiation of the function execution.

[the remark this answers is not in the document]
lukas016: ok we can solve it with docs

provider/cmd/app/app.go
so-sahu: replaced.

provider/cmd/app/app.go
so-sahu: renamed

lukas016 (commented):
lgtm i am waiting for rebase before approval

lukas016 (dismissed):

pkg/host/store.go
```diff
@@ -163,11 +163,6 @@ func (s *Store[E]) Delete(ctx context.Context, id string) error {
 		return fmt.Errorf("failed to set object metadata: %w", err)
 	}
 
-	s.enqueue(store.WatchEvent[E]{
-		Type:   store.WatchEventTypeDeleted,
-		Object: obj,
-	})
-
```
lukasfrank, line 170: Is there a reason for deleting this?

pkg/api/api.go
```diff
@@ -10,9 +10,10 @@ type Metadata struct {
 	Annotations map[string]string `json:"annotations"`
 	Labels      map[string]string `json:"labels"`
 
-	CreatedAt  time.Time  `json:"createdAt"`
-	DeletedAt  *time.Time `json:"deletedAt,omitempty"`
-	Generation int64      `json:"generation"`
+	CreatedAt       time.Time  `json:"createdAt"`
+	DeletedAt       *time.Time `json:"deletedAt,omitempty"`
+	FirstShutdownAt time.Time  `json:"firstShutdownAt"`
```
lukasfrank, line 15:
```suggestion
	ShutdownAt time.Time  `json:"ShutdownAt"`
```

pkg/controllers/machine_controller.go
```diff
@@ -235,6 +248,118 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
+			log.Error(err, "failed to list machines")
+			return
+		}
+
+		for _, machine := range machines {
+			if !slices.Contains(machine.Finalizers, MachineFinalizer) || machine.DeletedAt == nil {
+				continue
+			}
+
+			logger := log.WithValues("machineId", machine.ID)
```
lukasfrank, line 264:
```suggestion
			logger := log.WithValues("machineID", machine.ID)
```

lukasfrank (changes requested):

pkg/host/store.go
so-sahu: Since the garbage collector now fully manages the deletion process, there's no need to enqueue reconciliation. Setting the DeleteAt field alone is now adequate for gracefully shutting down the machine.

pkg/api/api.go
so-sahu: Presently, we are initiating machine shutdowns at regular intervals. However, in this field, we are preserving the timestamp when the initial shutdown attempt was made. Consequently, I considered that `FirstShutdownAt` might be a more fitting name for this field. Please share your thoughts on this.

pkg/controllers/machine_controller.go
so-sahu: Ok I will do the change.

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,108 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+			}
+		}
+
+	}, r.resyncIntervalGarbageCollector)
+}
+
+func (r *MachineReconciler) processMachineDeletion(ctx context.Context, log logr.Logger, machine *api.Machine, timeout time.Duration) error {
+	if !machine.Spec.ShutdownAt.IsZero() && time.Now().After(machine.Spec.ShutdownAt.Add(timeout)) {
+		err := r.destroyDomain(log, machine.ID)
+		if err != nil && !libvirtutils.IsErrorCode(err, libvirt.ErrNoDomain) {
```
lukas016, line 328:
```suggestion
		if err != nil && !libvirt.IsNotFound(err) {
```

lukas016 (changes requested):

pkg/controllers/machine_controller.go
so-sahu: Done

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,107 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+	if machine.Spec.ShutdownAt.Add(r.gcVMGracefulShutdownTimeout).Before(time.Now()) {
+		//Trigger machine shutdown until VMGracefulShutdownTimeout is over
+		return r.shutdownMachine(log, machine, domain)
+	}
+
+	return false, r.destroyDomain(log, domain)
+}
+
+func (r *MachineReconciler) destroyDomain(log logr.Logger, domain libvirt.Domain) error {
+	if err := r.libvirt.DomainDestroy(domain); err != nil {
```
lukas016, line 379: pls implement graceful destroy

so-sahu: done

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,107 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+		if libvirt.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to initiate forceful shutdown: %w", err)
+	}
+
+	log.V(1).Info("Triggered forceful shutdown")
```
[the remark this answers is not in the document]
so-sahu: changed it to `Destroyed domain`

lukas016 (dismissed):

lukasfrank (dismissed):

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,109 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+
+	if machine.Spec.ShutdownAt.IsZero() {
+		machine.Spec.ShutdownAt = time.Now()
+		if _, err := r.machines.Update(ctx, machine); err != nil {
+			return false, fmt.Errorf("failed to update ShutdownAt: %w", err)
+		}
+		log.V(1).Info("Updated machine ShutdownAt", "ShutdownAt", machine.Spec.ShutdownAt)
+	}
+
+	if time.Now().Before(machine.Spec.ShutdownAt.Add(r.gcVMGracefulShutdownTimeout)) {
```
afritzler, line 370: Won't we run the shutdown multiple times here until the shutdown time is reached? Is that the expected behavior?

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,109 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+			if err := r.processMachineDeletion(ctx, logger, machine); err != nil {
+				logger.Error(err, "failed to garbage collect machine")
+			}
+		}
+
+	}, r.resyncIntervalGarbageCollector)
+}
+
+func (r *MachineReconciler) processMachineDeletion(ctx context.Context, log logr.Logger, machine *api.Machine) error {
+	isDeleting, err := r.deleteMachine(ctx, log, machine)
```
afritzler, line 324: Can we add a log here stating the the deletion is triggered?

pkg/controllers/machine_controller.go
```diff
@@ -285,6 +298,109 @@ func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context
...
+
+func (r *MachineReconciler) shutdownMachine(log logr.Logger, machine *api.Machine, domain libvirt.Domain) (bool, error) {
+	log.V(1).Info("Triggering shutdown", "ShutdownAt", machine.Spec.ShutdownAt)
+	if err := r.libvirt.DomainShutdownFlags(domain, libvirt.DomainShutdownAcpiPowerBtn); err != nil {
+		if libvirt.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to initiate shutdown: %w", err)
+	}
+
```
afritzler, line 400: Add a log here that the `Shutdown` has been triggered.

afritzler (changes requested):
We might also want to add a test case ensuring the correct behavior e.g. start the service with a relatively short grace period.

pkg/controllers/machine_controller.go
lukas016: yes it is expected behavior because sometimes signal can be ignore and shutdown call is async call.

afritzler: How can a signal be "ignored"? Async should be not a problem here. I was just wondering why we should pounding a shutdown signal to the domain every X seconds.

lukas016: pls ask @gehoern, he recommended call it more times (min 2times).

lukas016: Async isn't problem and it help us here but it could be problem, if it is blocking call as Destroy.

pkg/controllers/machine_controller.go
lukas016: you have log before triggering. if it is end with error you will get error message. Isn't this better way how logs stuff? you know what was current operation and you don't need additional fields for errors.

so-sahu: Is it necessary in this context? Given that we are already logging `Triggering shutdown`, the absence of any subsequent error logs implies it's triggered.

pkg/controllers/machine_controller.go
so-sahu: We have the option to include it. However, this log will be reiterated in the console following every resync interval, in conjunction with the "Triggering shutdown" log. We already possess logs that clearly illustrate whether a shutdown or domain destruction has taken place. Is this the anticipated behavior?

so-sahu:
> We might also want to add a test case ensuring the correct behavior e.g. start the service with a relatively short grace period.

While the integration tests are currently in progress, it would be nice to include these tests there to prevent any potential conflicts.

pkg/controllers/machine_controller.go
lukasfrank: The `VM` can miss the `AcpiPowerBtn` signal in case of heavy load. Therefore the signal is sent multiple times. Maybe put a comment in the code to document it.

so-sahu: Comment added.

afritzler: Fair enought. Then I would suggest to put some comments in the code explaining this circumstances.

afritzler: I see this has been addressed already.

lukas016 (approved):

lukasfrank (approved):

afritzler (approved):

PR #103, opened 2024-01-09
improve setup logger v2 draft for discussion

lukas016:
It is just draft for discussion


lukasfrank (changes requested):
I think the proposed solution is not simple enough. What do you think about this alternative: https://github.com/ironcore-dev/libvirt-provider/pull/135

PR #104, opened 2024-01-10
Support for  `Volume` resize 

### LINKED ISSUE #58: Support for Resizing Attached Ceph Volumes
# Summary
This issue proposes a feature to enable resizing of already attached Ceph volumes in our system. The goal is to dynamically update the size of attached Ceph disks without manual intervention. 

Proposed Solution
The solution involves the development of a dedicated Go routine with the following functionalities:

1. Polling the Ceph Cluster: Continuously monitor the Ceph cluster to detect any changes in the size of the Ceph disks. This involves fetching the new size of the Ceph disk if it has been altered.  

2. Polling the VMs on the Host: Simultaneously, the routine should poll the virtual machines (VMs) running on the host to acquire the old-current size of the Ceph disk already attached to the VMs. The Go equivalent of the `virsh domblkinfo dom_uid vdb` command can be used.

3. Updating the Disk Size: When a discrepancy in size is detected, the routine should automatically update the size of the attached disk. This could be achieved with a command similar to `virsh blockresize dom_uid vdb 10G`.

## Motivation

Ensuring more flexibility and efficiency in disk management.

lukasfrank:
# Proposed Changes

- Added `Volume` resize support
- Added `GetSize` to the volume plugin interface

Fixes #58


pkg/api/machine.go
```diff
@@ -57,6 +57,7 @@ type VolumeStatus struct {
 	Name   string      `json:"name,omitempty"`
 	Handle string      `json:"handle,omitempty"`
 	State  VolumeState `json:"state,omitempty"`
+	Size   int64       `json:"size,omitempty"`
```
lukas016, line 60: Can be size negative?

pkg/controllers/machine_controller.go
```diff
@@ -57,13 +59,21 @@ var (
...
+	Host                     providerhost.Host
+	VolumePluginManager      *providervolume.PluginManager
+	NetworkInterfacePlugin   providernetworkinterface.Plugin
+	VolumeEvents             event.Source[*api.Machine]
+	ResyncDurationVolumeSize time.Duration
+}
+
+func setMachineReconcilerOptionsDefaults(o *MachineReconcilerOptions) {
+	if o.ResyncDurationVolumeSize == 0 {
+		o.ResyncDurationVolumeSize = time.Minute
```
lukas016, line 75: do you want to do it configurable? If yes, can you pls implement flag for it.

pkg/controllers/machine_controller.go
```diff
@@ -154,12 +169,53 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 		}
 	}()
 
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		wait.UntilWithContext(ctx, func(ctx context.Context) {
```
lukas016, line 177: don't use same name for context in anonymous function. because is extremely hard know on first look which context you use, because you can use closure principle too.

lukas016: anonymous function can be defined as standard separate function.

pkg/controllers/machine_controller.go
```diff
@@ -154,12 +169,53 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+			machines, err := r.machines.List(ctx)
+			if err != nil {
+				log.Error(err, "failed to list machines")
```
lukas016, line 180: optional: using of logger with own name can be useful for debuging

pkg/controllers/machine_controller.go
lukas016: Do you need such complex logic? Technically you calculate differences between volume sizes in every reconcile loop anyway. So why we cannot reconcile machine every 1min or 5min. How many VM do you want to run at the same time? Because our current limit is 128 vm per host

pkg/controllers/machine_controller_volumes.go
```diff
@@ -94,6 +95,22 @@ func (r *MachineReconciler) machineVolumeMounter(machine *api.Machine) VolumeMou
...
+func getVolumeStatus(machine *api.Machine, volumeID string) *api.VolumeStatus {
+	for _, volumeStatus := range machine.Status.VolumeStatus {
+		if volumeID == volumeStatus.Handle {
+			return &volumeStatus
+		}
+	}
+	return nil
+}
+
+func getLastVolumeSize(machine *api.Machine, volumeID string) *int64 {
```
lukas016, line 107: why do you return pointer? Do you need modified it? It can be dangerous.

pkg/plugins/volume/ceph/size.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/ceph/go-ceph/rados"
+	"github.com/ceph/go-ceph/rbd"
+	"github.com/ironcore-dev/libvirt-provider/pkg/api"
+)
+
+func createKeyFile(imageName string, key string) (string, func() error, error) {
+	noOpCleanup := func() error { return nil }
+	file, err := os.CreateTemp("", fmt.Sprintf("%s-*", imageName))
+	if err != nil {
+		return "", noOpCleanup, fmt.Errorf("failed to create temp file: %w", err)
```
lukas016, line 23: return "", nil, fmt.Errorf(...) this cannot work?

pkg/plugins/volume/ceph/size.go
```diff
@@ -0,0 +1,125 @@
...
+		return 0, fmt.Errorf("handle is not well formated: expected 'pool/image' format but got %s", spec.Connection.Handle)
+	}
+	poolName, imageName := parts[0], parts[1]
+
+	keyFile, cleanup, err := createKeyFile(imageName, userKey)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create temp key file: %w", err)
+	}
+	defer func() {
+		_ = cleanup()
```
lukas016, line 91: i think we can minimally log this error with name of file. it can be useful for cleanup of inactive files.

lukas016 (changes requested):

pkg/api/machine.go
lukasfrank: The size can't be negative but I'm consistent with the other size datatypes.

pkg/controllers/machine_controller.go
lukasfrank: Here we do only enqueue if the machine needs to be reconciled. It's an anti-pattern to just enqueue objects if you like to achieve a trigger to act on the object.

pkg/plugins/volume/ceph/size.go
lukasfrank: no, if you call a `nil` function you will see an exception.

pkg/controllers/machine_controller.go
lukas016: okay and what another 2 issues?

pkg/controllers/machine_controller.go
```diff
@@ -154,12 +169,54 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 		}
 	}()
 
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		sizeLogger := r.log.WithName("size")
+		defer wg.Done()
+
+		wait.UntilWithContext(ctx, func(ctx context.Context) {
```
lukas016, line 178: anonymous function can be defined as standard function separately.

lukas016, line 178: don't use same name for context in anonymous function. because is extremely hard know on first look which context you use, because you can use closure principle too.

pkg/controllers/machine_controller.go
lukas016: i created new issues for them

lukas016 (changes requested):

pkg/controllers/machine_controller.go
lukasfrank: Will keep it like this, since otherwise the logger needs to be created on every function call.

provider/cmd/app/app.go
```diff
@@ -97,6 +99,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.RootDir, "libvirt-provider-dir", filepath.Join(homeDir, ".libvirt-provider"), "Path to the directory libvirt-provider manages its content at.")
 
 	fs.StringVar(&o.PathSupportedMachineClasses, "supported-machine-classes", o.PathSupportedMachineClasses, "File containing supported machine classes.")
+	fs.DurationVar(&o.ResyncDurationVolumeSize, "volume-size-resync-duration", o.ResyncDurationVolumeSize, "Interval to determine volume size changes.")
```
afritzler, line 102: Shouldn't that be rather called interval instead of duration?

pkg/plugins/volume/ceph/size.go
```diff
@@ -0,0 +1,130 @@
...
+	cleanup := func() error {
+		return os.Remove(file.Name())
+	}
+
+	_, err = file.WriteString(key)
+	if err != nil {
+		return "", cleanup, fmt.Errorf("failed to write key to temp file: %w", err)
+	}
+
+	return file.Name(), cleanup, nil
```
afritzler, line 35: Where are those return `funcs` are being called? I assume you want to do a `defer cleanup`.

afritzler (changes requested):

pkg/plugins/volume/ceph/size.go
lukasfrank: Here: 
https://github.com/ironcore-dev/libvirt-provider/blob/1e6ff11e7cd27d426db68bc1f69ff28db919bcf3/pkg/plugins/volume/ceph/size.go#L93C2-L97C5
The authentication file needs to stay until the connection is being closed, therefore the cleanup needs to happen outside of this function.

pkg/plugins/volume/ceph/size.go
lukas016: but you don't call it. if you return err, you don't continue in execution. Do you want to keep is more general for use?

pkg/controllers/machine_controller.go
lukas016: you have 2 options:

- use combination of anonymous function with standard function and use closure principle for arguments.
- don't use wait but only simple two for loops.

And pls stop resolving conversation, i am author of conversation and i have to be ok with your fixes or arguments. Resolve conversation of someone else has to be edge case (sick, doesn't react, ...).

It is extremely hard check what you fixed or what you just closed. Especially when one conversion affects more conversation i cannot use outdated label and i have to open every conversantion.

provider/cmd/app/app.go
lukas016: maybe we can add additional information into help. It is interval between calling of function. So if check takes 5min (network issues), it can be confusing.

pkg/controllers/machine_controller.go
```diff
@@ -154,12 +169,54 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 		}
 	}()
 
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		sizeLogger := r.log.WithName("size")
```
lukas016, line 175:
```suggestion
		sizeLogger := r.log.WithName("volume-size-checker")
```

or something similar but size can be confusing for ops guys

pkg/controllers/machine_controller.go
```diff
@@ -590,7 +647,7 @@ func (r *MachineReconciler) setDomainMetadata(log logr.Logger, machine *api.Mach
 	return nil
 }
 
-func (r *MachineReconciler) setDomainResources(ctx context.Context, log logr.Logger, machine *api.Machine, domain *libvirtxml.Domain) error {
+func (r *MachineReconciler) setDomainResources(machine *api.Machine, domain *libvirtxml.Domain) error {
```
lukas016, line 650: FYI: it isn't very good practice doing cleaning changes with functional changes. I understand it is easier but It little complicates review process (this changes doesn't do anything with Issue) and depends on merge functionality you can have problem with reversing of commit.

pkg/controllers/machine_controller.go
```diff
@@ -154,12 +169,54 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+					}
+
+					volumeSize, err := plugin.GetSize(innerCtx, volume)
+					if err != nil {
+						sizeLogger.Error(err, "failed to get volume size", "machineID", machine, "volume", volume.Name)
+						continue
+					}
+
+					if lastVolumeSize := getLastVolumeSize(machine, volume.Name); volumeSize != ptr.Deref(lastVolumeSize, 0) {
+						sizeLogger.V(1).Info("Enqueue machine: Volume Size changed", "machineID", machine.ID, "lastSize", lastVolumeSize, "volumeSize", volumeSize)
```
lukas016, line 202: missing volume field with name?

pkg/controllers/machine_controller_volumes.go
```diff
@@ -130,17 +147,23 @@ func (r *MachineReconciler) attachDetachVolumes(ctx context.Context, log logr.Lo
...
-			volumeStates = append(volumeStates, api.VolumeStatus{
-				Name:   volume.Name,
-				Handle: volumeID,
-				State:  api.VolumeStateAttached,
-			})
+			continue
 		}
+
+		if lastVolumeSize := getLastVolumeSize(machine, volumeID); lastVolumeSize != nil && volumeSize != ptr.Deref(lastVolumeSize, 0) {
+			log.V(1).Info("Resizing volume", "lastSize", lastVolumeSize, "volumeSize", volumeSize)
```
lukas016, line 157: missing name of volume. It can be useful primary if you have multiple workers

pkg/controllers/machine_controller.go
lukasfrank: As already mentioned, I will keep it like this since it dosen't bring any benefits.

pkg/controllers/machine_controller.go
lukasfrank: `WithName` appends to the already given name which is `machine-reconciler`. I think `machine-reconciler-volume-size-checker` is too verbose.

provider/cmd/app/app.go
lukasfrank: > Shouldn't that be rather called interval instead of duration?

fixed the description to have it consistent

pkg/controllers/machine_controller.go
lukas016: you create spaghetti code and from ~100 lines function is 40 lines of code independent logic. If you create separate function as checkingVolumeSizeAsRequeue, it will be more readable and people don't have to study of code for know what it is doing.

pkg/controllers/machine_controller.go
lukas016: ok, i agree but size is still confusing because it can be size of vm (cpu, memory). Maybe volume-size?

provider/cmd/app/app.go
lukas016: Sorry but i prefer as @afritzler mention interval. Duration meaning resync function has to be done in specified time. It isn't semantic correctly, because this variable modified interval between 2 calls of resync function

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +194,44 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+			return
+		}
+
+		for _, machine := range machines {
+			var shouldEnqueue bool
+
+			for _, volume := range machine.Spec.Volumes {
+				plugin, err := r.volumePluginManager.FindPluginBySpec(volume)
+				if err != nil {
+					log.Error(err, "failed to get volume plugin", "machineID", machine, "volume", volume.Name)
```
lukas016, line 211: inconsistent key name volume (vs volumeName) with different part of code: https://github.com/ironcore-dev/libvirt-provider/pull/104/files#diff-7d2054388a9e1a4b7f89c284e697641903421c8fa3e63e30c15cacb9ab393979R157

Keeping of the same name can be useful for filtering of log records.

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +194,44 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+				}
+
+				volumeSize, err := plugin.GetSize(ctx, volume)
+				if err != nil {
+					log.Error(err, "failed to get volume size", "machineID", machine, "volume", volume.Name)
+					continue
+				}
+
+				if lastVolumeSize := getLastVolumeSize(machine, volume.Name); volumeSize != ptr.Deref(lastVolumeSize, 0) {
+					log.V(1).Info("Enqueue machine: Volume Size changed", "volume", volume.Name, "machineID", machine.ID, "lastSize", lastVolumeSize, "volumeSize", volumeSize)
```
lukas016, line 222: inconsistent key name volume (vs volumeName) with different part of code: https://github.com/ironcore-dev/libvirt-provider/pull/104/files#diff-7d2054388a9e1a4b7f89c284e697641903421c8fa3e63e30c15cacb9ab393979R157

Keeping of the same name can be useful for filtering of log records.

provider/cmd/app/app.go
lukas016: i am okay with current current fix. If you want to you can close issue, i am not author so i don't want to do that.

pkg/plugins/volume/ceph/size.go
lukas016: Are you sure it will contain name of file? because i don't think that err has to contain name of file.

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +194,44 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 	return nil
 }
 
+func (r *MachineReconciler) processVolumeSizeChanges(parent context.Context, log logr.Logger) {
```
afritzler, line 197: Can we use `ctx` instead of `parent` here.

afritzler, line 197: Can we call this method `startCheckAndEnqueueVolumeResize ` or `startMonitorVolumeForResizeReconcile` - or something like that.

pkg/plugins/volume/ceph/size.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/ironcore-dev/libvirt-provider/pkg/api"
+)
+
+func createKeyFile(imageName string, key string) (string, func() error, error) {
+	noOpCleanup := func() error { return nil }
+	file, err := os.CreateTemp("", fmt.Sprintf("%s-*", imageName))
+	if err != nil {
+		return "", noOpCleanup, fmt.Errorf("failed to create temp file: %w", err)
+	}
+	cleanup := func() error {
```
afritzler, line 26: Why not returning the file handler here and do the `defer os.RemoveFile(tempFile)` whenever it is needed.

afritzler (changes requested):

pkg/controllers/machine_controller.go
lukasfrank: It was before `ctx` but for @lukas016 it was confusing hence I changed it.

pkg/plugins/volume/ceph/size.go
lukasfrank: `%w` wraps the error

pkg/controllers/machine_controller.go
lukas016: before you had anonymous function in start function with another context with the same name. here you can keep ctx as @afritzler suggest because you don't have anymore problem with the same names.

pkg/plugins/volume/ceph/size.go
lukas016: do we need file handler here? it can do more problems with multiple goroutines we can easily hit file descriptor limits and rbd command require path to file not file descriptor/handler.

pkg/plugins/volume/ceph/size.go
lukas016: i don't see any wrapping anywhere in code but minimally os.Remove return ErrorPath which contains name of file so it is good.

pkg/plugins/volume/ceph/size.go
lukas016: or what do you mean file handler? File path or File structure?

lukasfrank: Actually, I would like to keep it like this since currently the function is self contained. Moving `defer os.RemoveFile(tempFile)` to the outside makes the caller to know that some cleanup is needed and in case of using the function at multiple places spreads this logic around the project.

pkg/controllers/machine_controller.go
lukasfrank: @lukas016 To be precise the name `ctx` is still overloaded as you can see here `wait.UntilWithContext(ctx, func(ctx context.Context) {` .

But anyhow before and now I prefer to name a context like `ctx`.

lukas016: @lukasfrank yes you have right, my fault. But in this case you have previous problem, what about innerCtx? Because it little looks as closure but it isn't true.

pkg/plugins/volume/ceph/size.go
lukas016: i agree with @lukasfrank, current solution looks fine for me.

afritzler (approved):

PR #113, opened 2024-01-16
Implemented hugepages support

### LINKED ISSUE #107: Add hugepages block in domain
# Summary

Libvirt provider doesn't support usage of hugepages for VM currently. It impacts VMs performance maybe around 30%.
Enable of hugepages is very simple, we need to add extra option into VM domain during creation process.
We can implement simple workaround for integrate support of hugepages into libvirt-provider for improve user experience.

It will be temporary workaround because correct using of hugepages has to be improve for using with cpu pinning and support of numa.

## Tasks

- [x] implement support for hugepages into libvirt-provider domain
  - it has to be optional functionality enabled with argument
- [x] implement additional logic for calculation of resources
  - we have to send size of hugepages and not whole memory

## Motivation

Improve VM performance easily and improve user experience

lukas016:
# Proposed Changes

- Implement basic hugepages support
- If enabled announce memory of hugepages instead 

Fixes #107

### DIFF
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -68,6 +68,7 @@ type MachineReconcilerOptions struct {
 	NetworkInterfacePlugin   providernetworkinterface.Plugin
 	VolumeEvents             event.Source[*api.Machine]
 	ResyncIntervalVolumeSize time.Duration
+	EnableHugepages          bool
 }
 
 func setMachineReconcilerOptionsDefaults(o *MachineReconcilerOptions) {
@@ -111,6 +112,7 @@ func NewMachineReconciler(
 		volumePluginManager:      opts.VolumePluginManager,
 		networkInterfacePlugin:   opts.NetworkInterfacePlugin,
 		resyncIntervalVolumeSize: opts.ResyncIntervalVolumeSize,
+		enableHugepages:          opts.EnableHugepages,
 	}, nil
 }
 
@@ -124,6 +126,7 @@ type MachineReconciler struct {
 	host              providerhost.Host
 	imageCache        providerimage.Cache
 	raw               raw.Raw
+	enableHugepages   bool
 
 	volumePluginManager    *providervolume.PluginManager
 	networkInterfacePlugin providernetworkinterface.Plugin
@@ -656,10 +659,17 @@ func (r *MachineReconciler) setDomainResources(machine *api.Machine, domain *lib
 		Unit:  "Byte",
 	}
 
+	if r.enableHugepages {
+		domain.MemoryBacking = &libvirtxml.DomainMemoryBacking{
+			MemoryHugePages: &libvirtxml.DomainMemoryHugepages{},
+		}
+	}
+
 	cpu := uint(machine.Spec.CpuMillis / 1000)
 	domain.VCPU = &libvirtxml.DomainVCPU{
 		Value: cpu,
 	}
+
 	return nil
 }
 
--- a/pkg/mcr/registry.go
+++ b/pkg/mcr/registry.go
@@ -76,15 +76,15 @@ func GetQuantity(class *iri.MachineClass, host *Host) int64 {
 	return int64(math.Min(float64(cpuRatio), float64(memoryRatio)))
 }
 
-func GetResources(ctx context.Context) (*Host, error) {
+func GetResources(ctx context.Context, enableHugepages bool) (*Host, error) {
 	hostMem, err := mem.VirtualMemoryWithContext(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get host memory: %w", err)
+		return nil, fmt.Errorf("failed to get host memory information: %w", err)
 	}
 
 	hostCPU, err := cpu.InfoWithContext(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get host memory: %w", err)
+		return nil, fmt.Errorf("failed to get host cpu information: %w", err)
 	}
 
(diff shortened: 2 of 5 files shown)


pkg/mcr/registry.go
```diff
@@ -36,23 +36,38 @@ func LoadMachineClassesFile(filename string) ([]iri.MachineClass, error) {
...
-		classes: map[string]iri.MachineClass{},
+		classes:          map[string]iri.MachineClass{},
+		disableHugepages: disableHugepages,
+	}
+
+	var hugepageSize int64
+	if !disableHugepages {
+		hostMem, err := mem.VirtualMemoryWithContext(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get host memory: %w", err)
```
so-sahu, line 49: Currently, error messages for `cpu.InfoWithContext` and `mem.VirtualMemoryWithContext` are the same. Could you please change the message for `cpu.InfoWithContext`.

pkg/mcr/registry.go
```diff
@@ -92,10 +111,26 @@ func GetResources(ctx context.Context) (*Host, error) {
...
+			host.Mem = resource.NewQuantity(int64(hostMem.HugePageSize*hostMem.HugePagesFree), resource.BinarySI)
+		}
+	}
+	return host, nil
+}
+
+func roundHugepagesUp(class *iri.MachineClass, hugepageSize int64) {
+	hugepageCount := int64(math.Ceil(float64(class.GetCapabilities().MemoryBytes) / float64(hugepageSize)))
+	class.GetCapabilities().MemoryBytes = hugepageCount * int64(hugepageSize)
+	fmt.Println(hugepageCount, hugepageSize, class.GetCapabilities().MemoryBytes)
```
so-sahu, line 133: `fmt.Println` can be removed.

lukas016: fixed

pkg/controllers/machine_controller.go
```diff
@@ -656,10 +659,17 @@ func (r *MachineReconciler) setDomainResources(machine *api.Machine, domain *lib
 		Unit:  "Byte",
 	}
 
+	if !r.disableHugepages {
+		domain.MemoryBacking = &libvirtxml.DomainMemoryBacking{
+			MemoryHugePages: &libvirtxml.DomainMemoryHugepages{},
+		}
+	}
+
```
lukasfrank, line 667: I would propose to keep only this for now until we handle the resources correctly. The other changes will break contracts and we announce incorrect resources.

provider/server/server.go
```diff
@@ -93,6 +94,7 @@ func (s *Server) loggerFrom(ctx context.Context, keysWithValues ...interface{})
 type MachineClassRegistry interface {
 	Get(volumeClassName string) (*iri.MachineClass, bool)
 	List() []*iri.MachineClass
+	GetResources(context.Context) (*mcr.Host, error)
```
lukasfrank, line 97: Please keep the interface simple until there is a need for this change. 
The  `GetResources` from `(m *Mcr)` does not have any important reference to `Mcr` which is not known outside of the function. 
```
func (m *Mcr) GetResources(ctx context.Context) (*Host, error) {
	return GetResources(ctx, m.disableHugepages)
}
```

lukasfrank (changes requested):

pkg/mcr/registry.go
```diff
@@ -36,23 +36,38 @@ func LoadMachineClassesFile(filename string) ([]iri.MachineClass, error) {
...
 	}
 
 	for _, class := range classes {
 		if _, ok := registry.classes[class.Name]; ok {
 			return nil, fmt.Errorf("multiple classes with same name (%s) found", class.Name)
 		}
+
+		if hugepageSize != 0 {
+			roundHugepagesUp(&class, hugepageSize)
+		}
```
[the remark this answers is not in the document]
lukas016: i think, you have very bad design.

pkg/controllers/machine_controller.go
lukas016: it can create wrong behavior because same machines will use hugepages other not. So you can have performance differences between same type of machine. What can be confusing for users. And i am not sure, if we can create machine without free hugepages, Adree mentioned something it is possible, but i didn't test it. But you will have problem with different performance.

pkg/mcr/registry.go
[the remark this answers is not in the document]
lukas016: after offline discussion and clarified some stuff, i agree design is good.

pkg/controllers/machine_controller.go
lukas016: we will have problem with different performance, i minimized problem with usage of total count of hugepages for calculate quantity but problem still exists.

provider/cmd/app/app.go
```diff
@@ -109,6 +111,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	flag.StringVar(&virshExecutable, "virsh-executable", "virsh", "Path / name of the virsh executable.")
 
+	fs.BoolVar(&o.DisableHugepages, "disable-hugepages", false, "Disable using Hugepages.")
```
lukasfrank, line 114: Can we please disable the usage of hugepages by default?

lukas016: it is possible. it is copied approach from virtlet.

lukas016: fixed

lukasfrank (approved):

so-sahu (approved):

PR #115, opened 2024-01-16
Reconcile `Machine`s  when `libvirt` VM state changes

### LINKED ISSUE #102: `CreateMachine()` always returns `PENDING` state for new Machines with No Disks/Volumes/Network
**Describe the bug**
Based on the current implementation, the initial reconciliation process for machine creation occurs in three stages:
- Stage 1: The finalizer is set in the API machine object.
- Stage 2: The libvirt machine is created, and pending machine state, volume, and network details are added to the API machine object.
- Stage 3: The actual machine state is updated in the API machine object.

However, there is an issue when attempting to create a machine with no disks and volumes. In such cases, the machine is not being requeued in Stage 2 due to no change in the API machine object. Consequently, Stage 3 reconciliation does not occur and the machine state is remaining `PENDING`.

**To Reproduce**
1. Attempt to create a machine with no disks and no network.
2. Wait for machine domain to be in `running` state - `watch virsh -c qemu:///system list`
3. The machine state still shows `PENDING` in the store.

**Expected behavior**
The machine state in the store must align with the state of the corresponding libvirt machine.

so-sahu:
Added functionality to periodically check the libvirt machine state and trigger a reconcile if there is a state mismatch. 

Fixes https://github.com/ironcore-dev/libvirt-provider/issues/102

### DIFF
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -56,6 +56,8 @@ var (
 		//libvirt.DomainShutoff:     api.MachineStateShutdown,
 		libvirt.DomainPmsuspended: api.MachineStatePending,
 	}
+
+	errNeedToRequeue = errors.New("need to requeue")
 )
 
 type MachineReconcilerOptions struct {
@@ -256,7 +258,9 @@ func (r *MachineReconciler) processNextWorkItem(ctx context.Context, log logr.Lo
 	ctx = logr.NewContext(ctx, log)
 
 	if err := r.reconcileMachine(ctx, id); err != nil {
-		log.Error(err, "failed to reconcile machine")
+		if err != errNeedToRequeue {
+			log.Error(err, "failed to reconcile machine")
+		}
 		r.queue.AddRateLimited(item)
 		return true
 	}
@@ -307,6 +311,9 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 	}
 	log.V(1).Info("Reconciled domain")
 
+	if state == api.MachineStatePending && nicStates == nil && volumeStates == nil {
+		return errNeedToRequeue
+	}
 	machine.Status.VolumeStatus = volumeStates
 	machine.Status.NetworkInterfaceStatus = nicStates
 	machine.Status.State = state
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -56,8 +56,6 @@ var (
 		//libvirt.DomainShutoff:     api.MachineStateShutdown,
 		libvirt.DomainPmsuspended: api.MachineStatePending,
 	}
-
-	errNeedToRequeue = errors.New("need to requeue")
 )
 
 type MachineReconcilerOptions struct {
@@ -71,6 +69,7 @@ type MachineReconcilerOptions struct {
 	VolumeEvents             event.Source[*api.Machine]
 	ResyncIntervalVolumeSize time.Duration
 	EnableHugepages          bool
+	MachineStateSyncInterval time.Duration
 }
 
 func setMachineReconcilerOptionsDefaults(o *MachineReconcilerOptions) {
@@ -115,6 +114,7 @@ func NewMachineReconciler(
 		networkInterfacePlugin:   opts.NetworkInterfacePlugin,
 		resyncIntervalVolumeSize: opts.ResyncIntervalVolumeSize,
 		enableHugepages:          opts.EnableHugepages,
+		machineStateSyncInterval: opts.MachineStateSyncInterval,
 	}, nil
 }
 
@@ -137,6 +137,8 @@ type MachineReconciler struct {
 	machineEvents event.Source[*api.Machine]
 
 	resyncIntervalVolumeSize time.Duration
+
+	machineStateSyncInterval time.Duration
 }
 
 func (r *MachineReconciler) Start(ctx context.Context) error {
@@ -179,6 +181,12 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 	go func() {
 		defer wg.Done()
 		r.startCheckAndEnqueueVolumeResize(ctx, r.log.WithName("volume-size"))
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		r.startCheckAndEnqueueMachineStatus(ctx, r.log.WithName("machine-status-sync"))
 	}()
 
 	go func() {
@@ -246,6 +254,33 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
 	}, r.resyncIntervalVolumeSize)
 }
 
+func (r *MachineReconciler) startCheckAndEnqueueMachineStatus(ctx context.Context, log logr.Logger) {
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
+		machines, err := r.machines.List(ctx)
+		if err != nil {
+			log.Error(err, "failed to list machines")
+			return
+		}
+
+		for _, machine := range machines {
+			if machine.DeletedAt != nil || !slices.Contains(machine.Finalizers, MachineFinalizer) {
+				continue
+			}
+			id := machine.ID
+
+			state, err := r.getMachineState(id)
+			if err != nil {
+				log.Error(err, "failed to get machine state", "machine:", id)
+				continue
+			}
+
+			if state != machine.Status.State {
+				r.queue.AddRateLimited(id)
+			}
+		}
+	}, r.machineStateSyncInterval)
+}
+
 func (r *MachineReconciler) processNextWorkItem(ctx context.Context, log logr.Logger) bool {
 	item, shutdown := r.queue.Get()
 	if shutdown {
@@ -258,9 +293,7 @@ func (r *MachineReconciler) processNextWorkItem(ctx context.Context, log logr.Lo
 	ctx = logr.NewContext(ctx, log)
 
 	if err := r.reconcileMachine(ctx, id); err != nil {
-		if err != errNeedToRequeue {
-			log.Error(err, "failed to reconcile machine")
-		}
+		log.Error(err, "failed to reconcile machine")
 		r.queue.AddRateLimited(item)
 		return true
 	}
@@ -311,9 +344,6 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 	}
 	log.V(1).Info("Reconciled domain")
 
-	if state == api.MachineStatePending && nicStates == nil && volumeStates == nil {
-		return errNeedToRequeue
-	}
 	machine.Status.VolumeStatus = volumeStates
 	machine.Status.NetworkInterfaceStatus = nicStates
 	machine.Status.State = state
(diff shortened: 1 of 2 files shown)


Rohit-0505:
@so-sahu I would recommend to add corresponding test cases also to validate the correct creation of machine with no **Network/Volume**

pkg/controllers/machine_controller.go
```diff
@@ -236,6 +240,9 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 	}
 	log.V(1).Info("Reconciled domain")
 
+	if state == api.MachineStatePending && nicStates == nil && volumeStates == nil {
```
lukas016, line 243: do you need this volumeStates or nicStates?

so-sahu: yes, otherwise, if a machine transitions from, let's say, 'Running' to 'Pending,' the reconciliation process will occur without the updated state being accurately reflected in the store.

so-sahu:
> @so-sahu I would recommend to add corresponding test cases also to validate the correct creation of machine with no **Network/Volume**

@Rohit-0505 As the integration tests are currently underway, it is recommended to incorporate this to prevent potential conflicts.

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +185,33 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+
+		for _, machine := range machines {
+			if machine.DeletedAt != nil || !slices.Contains(machine.Finalizers, MachineFinalizer) {
+				continue
+			}
+			id := machine.ID
+
+			state, err := r.getMachineState(id)
+			if err != nil {
+				log.Error(err, "failed to get machine state", "machine:", id)
```
lukas016, line 204: key is without : and pls unified name of key to machineId

pkg/controllers/machine_controller.go
```diff
@@ -173,6 +185,33 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
...
+				continue
+			}
+
+			if state != machine.Status.State {
```
lukas016, line 208: how solve it your issue with stucked domain reconcilation loop

lukas016 (changes requested):

pkg/controllers/machine_controller.go
lukas016: doesn't matter anymore

pkg/controllers/machine_controller.go
so-sahu: In store, the machine state will be "Pending". The synchronizer will get the state as "Running" and put it to the queue. (This is where it solves the stuck domain reconciliation loop)
In the next reconciliation, the correct machine state will be set in the end.

I have tested it and working fine.

pkg/controllers/machine_controller.go
so-sahu: Could you be please more clear on this?

lukas016:
```suggestion
				log.Error(err, "failed to get machine state", "machineId", id)
```

lukas016: i don't see more log message now. so i don't see any benefits in create new logger.

afritzler (dismissed):

afritzler:
PR looks good to me. Could you please rebase your PR branch?

so-sahu:
Rebase done.

afritzler (dismissed):

lukas016 (dismissed):

pkg/controllers/machine_controller.go
```diff
@@ -235,6 +245,33 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
+
+		for _, machine := range machines {
+			if machine.DeletedAt != nil || !slices.Contains(machine.Finalizers, MachineFinalizer) {
+				continue
+			}
+			id := machine.ID
+
+			state, err := r.getMachineState(id)
+			if err != nil {
+				log.Error(err, "failed to get machine state", "machineID", id)
```
hardikdr, line 264: If the libvirt domain is deleted from the host by-hand/mistakenly, the error would be ignored here. 
If that case is not handled somewhere else, it would be worth re-queueing the machine here, to have it back again.

hardikdr (commented):
If the libvirt domain is deleted from the host by-hand/mistakenly, the error would be ignored here.
If that case is not handled somewhere else, it would be worth re-queueing the machine here, to have it back again.

pkg/controllers/machine_controller.go
so-sahu: In the current design of the libvirt provider, we haven't accounted for manual intervention. In this situation, simply recreating the domain won't be adequate. We must also verify that the data and network configurations match those of the previous machine. Should we address such scenarios in this pull request?

hardikdr: Imo if you re-enqueue the machine if `libvirt.IsNotFound(err)` is seen (as the error is from libvirt client), the Reconcile method should create an intended fresh machine. The network and volume config should remain the same.

so-sahu: Agreed.
 However, it's important to note that data stored on an ephemeral disk is temporary and will be lost. Additionally, is there any practical scenario in a production environment where such a situation could occur?

lukasfrank: I think we should not  optimise for such scenarios. If a human operator fiddles around on the host he should anyway know what he is doing.

pkg/controllers/machine_controller.go
```diff
@@ -135,6 +137,8 @@ type MachineReconciler struct {
 	machineEvents event.Source[*api.Machine]
 
 	resyncIntervalVolumeSize time.Duration
+
+	machineStateSyncInterval time.Duration
```
lukasfrank, line 141:
```suggestion
	resyncIntervalMachineState time.Duration
```

provider/cmd/app/app.go
```diff
@@ -124,6 +125,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.Libvirt.Qcow2Type, "qcow2-type", qcow2.Default(), fmt.Sprintf("qcow2 implementation to use. Available: %v", qcow2.Available()))
 
+	fs.DurationVar(&o.MachineStateSyncInterval, "machine-state-sync-interval", 1*time.Minute, "Synchronization interval for aligning local machine state with corresponding libvirt machine state.")
```
lukasfrank, line 128: maybe move this below 
```
fs.DurationVar(&o.ResyncIntervalVolumeSize, "volume-size-resync-duration", o.ResyncIntervalVolumeSize, "Interval to determine volume size changes.")
```

provider/cmd/app/app.go
```diff
@@ -305,6 +308,7 @@ func Run(ctx context.Context, opts Options) error {
 			NetworkInterfacePlugin:   nicPlugin,
 			ResyncIntervalVolumeSize: opts.ResyncIntervalVolumeSize,
 			EnableHugepages:          opts.EnableHugepages,
+			MachineStateSyncInterval: opts.MachineStateSyncInterval,
```
lukasfrank, line 311: please keep it consistent with `	ResyncIntervalVolumeSize: opts.ResyncIntervalVolumeSize,`

lukasfrank (dismissed):

pkg/controllers/machine_controller.go
lukasfrank: Looking at the issue from the perspective that it not necessarily mean if the VM is not there that someone deleted it manually I also agree with the comment.

hardikdr: >However, it's important to note that data stored on an ephemeral disk is temporary and will be lost.

As the naming of the Ephemeral Disk is generated from the MachineID(which remains same) instead of the DomainUUID, ephe=disks should remain as they are, if only VM/Domain is missing/deleted.
```
	diskFilename := p.diskFilename(spec.Name, machine.ID)
```
(please verify the behavior though)

Ephemeral Disks would have been cleaned up only if the controller decided to clean up, which is desirable anyways. 

I am not super keen on necessarily fixing the issue in this PR, just food for thought from the use case pov :))

so-sahu: ok. Makes sense. I will log and re-enqueue the machine if `libvirt.IsNotFound(err)` is seen. Thanks for the clarity.

pkg/controllers/machine_controller.go
```diff
@@ -235,6 +244,33 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
+
+		for _, machine := range machines {
+			if machine.DeletedAt != nil || !slices.Contains(machine.Finalizers, MachineFinalizer) {
+				continue
+			}
+			id := machine.ID
+
+			state, err := r.getMachineState(id)
+			if err != nil {
+				log.Error(err, "failed to get machine state", "machineID", id)
```
lukas016, line 263:
```suggestion
				log.Error(err, "failed to get machine state", "machineId", id)
```

Pls keep same keys between logs. Filtering can be case sensitive

lukas016 (changes requested):
inconsistent logging keys

provider/cmd/app/app.go
lukas016: we created inconsistency here with gc.

lukasfrank: graceful shutdown isn't in main therefore I would propose to address it in the graceful shutdown PR

pkg/controllers/machine_controller.go
lukasfrank: @lukas016 Don't you think it's better to be consistent? Across the entire project we use `machineID` (for logging and as var) why should we change it here?

lukas016: no problem, but pls create issue for unified https://github.com/ironcore-dev/libvirt-provider/blob/main/pkg/controllers/machine_controller.go#L246

lukas016: just this one is different. So we can fix it in this PR and we don't need issue. What do you think @lukasfrank

provider/cmd/app/app.go
lukas016: okay. But i see sourav didn't implement my recommendation. You have different format for variable and arguments. arguments follow pattern general -> specific. Here you follow pattern specific -> general. So you have reverted logic. It looks for me as complication as help.

This is different:
o.ResyncIntervalMachineState, "machine-state-resync-duration"

General -> specific has benefit because you can do grouping of variable base on name which serve for the same logic and you know they have some relationship. It is better extendable in future too.

If you don't want it, you don't have to do implement it.

pkg/controllers/machine_controller.go
lukasfrank: & here  https://github.com/ironcore-dev/libvirt-provider/blob/main/pkg/controllers/machine_controller.go#L211-L229
and the entire variable naming

provider/cmd/app/app.go
so-sahu: Following our offline discussion, I have updated the key to `ResyncIntervalMachineState`. We can handle the consistent naming convention separately.

pkg/controllers/machine_controller.go
so-sahu: Following our offline discussion, I've modified the key [here](https://github.com/ironcore-dev/libvirt-provider/blob/864eba2245b1606aa06f2a684e624bb32a33a6c8/pkg/controllers/machine_controller.go#L246) to `machineID`.

lukasfrank (dismissed):

lukas016 (approved):

lukasfrank (approved):

PR #121, opened 2024-01-17
Fix RBAC: allow `machinepoollet` to update NIC status

afritzler:
# Proposed Changes

- Fix RBAC: allow `machinepoollet` to update NIC status
- Propagate `NetworkInterface` IPs to the provider status
- Update integration tests to reflect the `Volume` creation


hardikdr (dismissed):

so-sahu (dismissed):

pkg/plugins/networkinterface/apinet/apinet.go
```diff
@@ -59,10 +60,10 @@ func (p *Plugin) Init(host providerhost.Host) error {
 	return nil
 }
 
-func onmetalIPsToAPInetIPs(ips []string) []net.IP {
-	res := make([]net.IP, len(ips))
+func ironcoreIPsToAPInetIPs(ips []string) []apinet.IP {
+	res := make([]apinet.IP, len(ips))
 	for i, ip := range ips {
-		res[i] = net.MustParseIP(ip)
+		res[i] = apinet.MustParseIP(ip)
```
lukas016, line 66: why MustParseIP? usage of panic and handler recovery looks more as error exception system. Proper way for golang is return err.

pkg/plugins/networkinterface/apinet/apinet.go
```diff
@@ -185,6 +186,11 @@ func (p *Plugin) Apply(ctx context.Context, spec *api.NetworkInterfaceSpec, mach
 	}
 
 	log.V(1).Info("Host device is ready", "HostDevice", hostDev)
+	var nicIPs []net.IP
```
lukas016, line 189: you can preallocate this array on the same size as len(apinetNic.Spec.IPs)

provider/server/machine.go
```diff
@@ -138,10 +138,15 @@ func (s *Server) getIRINICStatus(machine *api.Machine) ([]*iri.NetworkInterfaceS
 			return nil, fmt.Errorf("failed to get nic state: %w", err)
 		}
 
+		var nicIPs []string
```
lukas016, line 141: you can preallocate this array

pkg/plugins/networkinterface/apinet/apinet.go
afritzler: Fair point, but I would resolve this in a seperate PR as I want to change as least amount of logic as possible here.

lukasfrank (approved):

PR #135, opened 2024-01-22
Add logger to `main` package and improve logging

### LINKED ISSUE #85: improve app logging

lukasfrank:
# Proposed Changes

Improved logging
- Replaced main error logging from `fmt.Println` to logger
- Removed redundant logging by logger and cobra
- Disabled usage help once the flags are parsed 

Fixes #85

### DIFF
--- a/provider/cmd/app/app.go
+++ b/provider/cmd/app/app.go
@@ -142,6 +142,10 @@ func Command() *cobra.Command {
 			cmd.SetContext(ctrl.LoggerInto(cmd.Context(), ctrl.Log))
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			//flag parsing is done therefore we can silence the usage message
+			cmd.SilenceUsage = true
+			//error logging is done in the main
+			cmd.SilenceErrors = true
 			return Run(cmd.Context(), opts)
 		},
 	}
@@ -242,7 +246,7 @@ func Run(ctx context.Context, opts Options) error {
 		ceph.NewPlugin(),
 		emptydisk.NewPlugin(qcow2Inst, rawInst),
 	}); err != nil {
-		return fmt.Errorf("failed to initialize machine class registry: %w", err)
+		return fmt.Errorf("failed to initialize volume plugin manager: %w", err)
 	}
 
 	nicPlugin, nicPluginCleanup, err := opts.NicPlugin.NetworkInterfacePlugin()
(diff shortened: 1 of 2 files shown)


provider/cmd/main.go
```diff
@@ -13,9 +12,10 @@ import (
 
 func main() {
 	ctx := ctrl.SetupSignalHandler()
+	mainLog := ctrl.Log.WithName("main")
```
afritzler, line 15: Just use `log` here as it is clear from the context what this logger is being used for.

afritzler (changes requested):

provider/cmd/main.go
lukasfrank: fixed

afritzler (dismissed):

provider/cmd/app/app.go
```diff
@@ -372,7 +391,7 @@ func runGRPCServer(ctx context.Context, setupLog logr.Logger, log logr.Logger, s
 
 	grpcSrv := grpc.NewServer(
 		grpc.ChainUnaryInterceptor(
-			commongrpc.InjectLogger(log),
+			commongrpc.InjectLogger(log.WithName("grpc-server")),
```
lukas016, line 394: optional: why grpc? it is iri protocol server, which implement your iri protocol. grpc is used technology.

provider/cmd/main.go
```diff
@@ -13,9 +12,10 @@ import (
 
 func main() {
 	ctx := ctrl.SetupSignalHandler()
+	log := ctrl.Log.WithName("main")
 
 	if err := app.Command().ExecuteContext(ctx); err != nil {
-		fmt.Println(err.Error())
+		log.V(1).Error(err, "Error running libvirt provider")
```
lukas016, line 18:
```suggestion
		log.V(1).Error(err, "error running libvirt provider")
```

i prefer same format of error messages.

provider/cmd/app/app.go
lukasfrank: fixed

provider/cmd/main.go
lukasfrank: fixed

lukas016 (approved):

PR #139, opened 2024-01-24
Fixed `nil pointer` dereference 

### LINKED ISSUE #134: Redundant Assignment and Potential Nil Pointer Dereference in `reconcileDesiredNetworkInterface`
In the `reconcileDesiredNetworkInterface` method, the line `mountedNic.networkInterface.Handle = providerNic.Handle` appears redundant as it is not utilized. This assumes `mountedNic` is non-nil before checking the map element, which may cause a panic.

lukasfrank:
# Proposed Changes

- Fix usage of `mountedNic` before checking existence 

Fixes #134

### DIFF
--- a/pkg/controllers/machine_controller_nics.go
+++ b/pkg/controllers/machine_controller_nics.go
@@ -216,11 +216,12 @@ func (r *MachineReconciler) reconcileDesiredNetworkInterface(
 	}
 
 	mountedNic, ok := mountedNics[nic.Name]
-	mountedNic.networkInterface.Handle = providerNic.Handle
-	if ok && reflect.DeepEqual(mountedNic.networkInterface, providerNic) {
-		return &mountedNic, nil
-	}
 	if ok {
+		mountedNic.networkInterface.Handle = providerNic.Handle
+		if reflect.DeepEqual(mountedNic.networkInterface, providerNic) {
+			return &mountedNic, nil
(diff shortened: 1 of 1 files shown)


so-sahu (dismissed):

pkg/controllers/machine_controller_nics.go
```diff
@@ -216,7 +216,6 @@ func (r *MachineReconciler) reconcileDesiredNetworkInterface(
 	}
 
 	mountedNic, ok := mountedNics[nic.Name]
-	mountedNic.networkInterface.Handle = providerNic.Handle
 	if ok && reflect.DeepEqual(mountedNic.networkInterface, providerNic) {
```
hardikdr, line 219: Afai remember, this line was added while adapting the controllers to the new APINet resources with Axel, to prevent the DeepEqual from failing. 
This issue emerged as the `.Handle` was newly defined with APINet resource adaptations from the Plugin, but it is not set in the `MountedNic` structure. 
I would recommend testing the flow with this change, and if it's still failing, maybe have a more holistic solution.

hardikdr (approved):

pkg/controllers/machine_controller_nics.go
hardikdr: Sure, thanks.

PR #144, opened 2024-01-29
update volume status after every reconcile loop

### LINKED ISSUE #114: `DetachVolume` failure: `Disk` is not getting unplugged from `Domain`
**Describe the bug**
Disk is not getting unplugged from domain when `DetachVolume()` IRI call is triggered. Volume gets removed from store but not from DomainXML.

**To Reproduce**
1. Create a machine with 2 empty disks
2. Detach one emty disk from machine
3. Verify domain XML for machine - empty disk will not get removed

**Expected behavior**
Empty disk should be removed from Domain XML also.

### LINKED ISSUE #145: empty-disk files are deleted before detach volumes from domain
**Describe the bug**
Raw files representing empty-disk can be deleted from FS before they are really detach from domain. Some problem can be with ceph volumes.
Root cause is wrong logic, where unmount (deletion for empty-disk) follow immediately after detach of volume call, if it doesn't return error. Problem is detach is async call (depends on hypervisor) and current state has to be check in domain description.

**To Reproduce**
1. Create unbootable vm (it is simpler for simulation) with multiple empty disks
2. Call detach for one disk
3. Check file on FS (It will be missing)

**Expected behavior**
Disks can be unmounted (deleted), when they are detached successfully.

lukas016:
# Proposed Changes

Rewrite detach/attach logic for using of status as source of truth.

- status of volumes is updated after every reconcile loop
- volumes are deleted from status, when they aren't part of domain anymore
- update state for detaching volumes

Fixes #114 
Fixes #145


pkg/controllers/machine_controller.go
```diff
@@ -343,12 +343,16 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 
 	log.V(1).Info("Reconciling domain")
 	state, volumeStates, nicStates, err := r.reconcileDomain(ctx, log, machine)
+
+	machine.Status.VolumeStatus = volumeStates
 	if err != nil {
+		if _, locErr := r.machines.Update(ctx, machine); locErr != nil {
+			log.Error(locErr, "failed to update image metadate")
```
lukas016, line 350: wrong message

pkg/api/machine.go
```diff
@@ -32,6 +32,20 @@ type MachineStatus struct {
...
 }
 
+func (m *MachineStatus) GetVolumesAsMap() map[string]*VolumeStatus {
+	if m == nil {
+		return map[string]*VolumeStatus{}
+	}
+
+	result := make(map[string]*VolumeStatus, len(m.VolumeStatus))
+	for index := range m.VolumeStatus {
+		// TODO: Using name isn't good idea because we cannot create volumes with the same name
```
lukas016, line 42: delete this comment, it isn't anymore problems.

pkg/controllers/machine_controller.go

pkg/api/machine.go
lukas016: fixed

pkg/controllers/machine_controller.go
```diff
@@ -456,17 +456,21 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 
 	log.V(1).Info("Reconciling domain")
 	state, volumeStates, nicStates, err := r.reconcileDomain(ctx, log, machine)
+
+	machine.Status.VolumeStatus = volumeStates
 	if err != nil {
+		if _, locErr := r.machines.Update(ctx, machine); locErr != nil {
```
so-sahu, line 462: Does it still make sense to proceed with the update even if volumeStates is nil?

lukas016: i don't think volumeStatus can be nil. I tried fixed every return values for that case.

so-sahu: https://github.com/ironcore-dev/libvirt-provider/blob/b7126900194d52d4d6debcd233f4d1134f520f8b/pkg/controllers/machine_controller.go#L487
Here, it can be nil.

lukas016: fixed

so-sahu: it can be nil, even in below:
- https://github.com/ironcore-dev/libvirt-provider/blob/c5809e04a24f80fd1d449af792cc9334b2989e3c/pkg/controllers/machine_controller.go#L521
- https://github.com/ironcore-dev/libvirt-provider/blob/c5809e04a24f80fd1d449af792cc9334b2989e3c/pkg/controllers/machine_controller.go#L526
- https://github.com/ironcore-dev/libvirt-provider/blob/c5809e04a24f80fd1d449af792cc9334b2989e3c/pkg/controllers/machine_controller.go#L561
- https://github.com/ironcore-dev/libvirt-provider/blob/c5809e04a24f80fd1d449af792cc9334b2989e3c/pkg/controllers/machine_controller.go#L566
- https://github.com/ironcore-dev/libvirt-provider/blob/c5809e04a24f80fd1d449af792cc9334b2989e3c/pkg/controllers/machine_controller.go#L572

I suggest to update when the volumeStatus is not nil.

pkg/controllers/machine_controller_volumes.go
```diff
@@ -83,75 +84,153 @@ func getLastVolumeSize(machine *api.Machine, volumeID string) *int64 {
...
+
+func convertVolumesMapToListAndNormalize(currentStatus map[string]*api.VolumeStatus) []api.VolumeStatus {
+	newVolumeStatus := make([]api.VolumeStatus, 0, len(currentStatus))
+	for _, status := range currentStatus {
+		newVolumeStatus = append(newVolumeStatus, *status)
+	}
+
+	// sort is required for avoid reconcile loop triggered by different order of volumes.
+	sort.SliceStable(newVolumeStatus, func(i, j int) bool {
+		return strings.Compare(newVolumeStatus[i].Name, newVolumeStatus[j].Name) == -1
```
so-sahu, line 230: Could you please explain why we are not simply using `newVolumeStatus[i].Name < newVolumeStatus[j].Name`. Both cases perform lexicographic comparison if I am correct.

so-sahu (changes requested):

pkg/controllers/machine_controller.go
```diff
@@ -456,17 +456,21 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
...
+			log.Error(locErr, "failed to update API machine state")
```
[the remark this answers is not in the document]
lukas016: volume status is current PR in future you can update network interfaces too. So i prefer keep it general

pkg/controllers/machine_controller.go
```diff
@@ -456,17 +456,21 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
...
 	}
 	log.V(1).Info("Reconciled domain")
 
-	machine.Status.VolumeStatus = volumeStates
 	machine.Status.NetworkInterfaceStatus = nicStates
 	machine.Status.State = state
 
 	if _, err = r.machines.Update(ctx, machine); err != nil {
-		return fmt.Errorf("failed to update image metadate: %w", err)
+		return fmt.Errorf("failed to update API machine state: %w", err)
```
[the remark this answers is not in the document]
lukas016: i will delete state word.

pkg/api/machine.go
```diff
@@ -37,6 +37,19 @@ type MachineStatus struct {
 	ImageRef               string                   `json:"imageRef"`
 }
 
+func (m *MachineStatus) GetVolumesAsMap() map[string]*VolumeStatus {
+	if m == nil {
```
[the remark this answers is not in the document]
lukas016: i cannot access to nil pointer in m.VolumeStatus, it will throw panic

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
lukas016: i will delete word state

pkg/api/machine.go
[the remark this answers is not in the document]
so-sahu: As far as I know, whether m is a pointer or not, when it is nil, accessing m.VolumeStatus will not cause a panic because nil slices in Go are safe to access. Could you please validate this once?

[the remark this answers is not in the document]
lukas016: you have right. but m isn't slice. m is pointer to struct which can be nil and i do access to field of struct which is array.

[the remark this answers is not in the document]
so-sahu: make sense. Thank you for the confirmation.

pkg/controllers/machine_controller.go
lukas016: Ok i will create function for it

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
lukas016: fixed

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
lukas016: fixed

pkg/controllers/machine_controller_volumes.go
lukas016: fixed

pkg/controllers/machine_controller.go
lukas016: fixed

pkg/controllers/machine_controller_volumes.go
```diff
@@ -83,75 +84,153 @@ func getLastVolumeSize(machine *api.Machine, volumeID string) *int64 {
...
-		return true
-	}); err != nil {
-		return nil, fmt.Errorf("error iterating attached volumes: %w", err)
+	currentStatus := machine.Status.GetVolumesAsMap()
+
+	attachedVolumes, err := attacher.ListVolumesAsMap()
+	if err != nil {
+		// missing list of attached volumes can affect deletion of volumes, so we cannot continue
+		return machine.Status.VolumeStatus, err
 	}
```
lukasfrank, line 112: I think it's a bad practice returning  "data" in an error case.

lukasfrank (changes requested):

pkg/controllers/machine_controller_volumes.go
```diff
@@ -83,75 +84,153 @@ func getLastVolumeSize(machine *api.Machine, volumeID string) *int64 {
...
 
-	if err := mounter.ForEachVolume(func(volume *MountVolume) bool {
-		currentVolumeNames.Insert(volume.ComputeVolumeName)
-		return true
-	}); err != nil {
-		return nil, fmt.Errorf("error iterating mounted volumes: %w", err)
+	syncStatusWithAttachedVolumes(currentStatus, attachedVolumes)
+
+	errs := r.deleteDetachedVolumes(ctx, log, currentStatus, mounter)
```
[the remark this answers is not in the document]
lukas016: functions are designed as failure tolerant. You have multiple logic on one place where one functionality can affect another (detach, delete, resize, attach). So one problematic disk cannot affect another.

pkg/controllers/machine_controller_volumes.go
lukas016: here we can return nil, it isn't problem with new update status functionality.

pkg/controllers/machine_controller.go
```diff
@@ -456,17 +456,18 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
...
 	log.V(1).Info("Reconciling domain")
 	state, volumeStates, nicStates, err := r.reconcileDomain(ctx, log, machine)
+
 	if err != nil {
+		locErr := r.updateAPIMachineStatus(ctx, machine, state, volumeStates, nicStates)
+		if locErr != nil {
+			log.Error(locErr, "failed to update API machine")
+		}
 		return providerimage.IgnoreImagePulling(err)
 	}
```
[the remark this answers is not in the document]
lukas016: if updateAPIMachineStatus fails and err is ImagePulling error, it will reconcile based on trigger from providerimage, which will be thrown when image will available (of course you don't have current status but machine domain doesn't exist in this time). It is same logic as before. If i return error base on updateAPIMachineStatus with ImagePulling error, we will end in loop, where we will wait for download of image. If err is different as ImagePulling, machine will reconcile again and status can be updated in next reconcile loop.

What do you suggest do here?

afritzler (changes requested):
I would suggest to subscribe to the `LifeCycleEvents` of the libvirt Domain and enqueue Machines if something changes in the domain.

Here is some sample code how this might look like:

```
// Subscribe to lifecycle events using the provided context
eventChan, err := l.LifecycleEvents(ctx)
if err != nil {
	fmt.Println("Failed to subscribe to lifecycle events:", err)
	return
}

fmt.Println("Subscribed to lifecycle events. Listening for events...")

// Listen for lifecycle events in a separate goroutine
go func() {
	for event := range eventChan {
		fmt.Printf("Received lifecycle event: %+v\n", event)
                // enqueue Machines here and switch over the events
	}
	fmt.Println("Lifecycle event channel closed.")
}()

// Wait for context to be cancelled (e.g., by receiving a signal)
<-ctx.Done()
fmt.Println("Context cancelled. Shutting down...")
```

This can be also used to consolidate other `go routines` which we are starting during the initialization of the provider.

lukas016:
@afritzler we discussed implement listen to libvirt events. And we want to do that but not now.

pkg/controllers/machine_controller.go
```diff
@@ -892,6 +893,33 @@ func (r *MachineReconciler) getDomainDesc(machineID string) (*libvirtxml.Domain,
...
+
+	if len(volumes) != 0 {
+		requireUpdate = true
+		machine.Status.VolumeStatus = volumes
+	}
+
+	if len(nics) != 0 {
+		requireUpdate = true
+		machine.Status.NetworkInterfaceStatus = nics
+	}
```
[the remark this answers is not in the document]
lukas016: fixed

pkg/controllers/machine_controller_volumes.go
lukas016: fixed

pkg/controllers/machine_controller.go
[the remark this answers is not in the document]
lukas016: i kept it as it is.

pkg/controllers/machine_controller_volumes.go
```diff
@@ -291,14 +372,44 @@ func (a *libvirtVolumeAttacher) diskByVolumeNameIndex(name string) (int, error)
...
+func (a *libvirtVolumeAttacher) loadVolumes() error {
+	if a.volumes != nil {
+		return nil
+	}
+
+	a.volumes = map[string]*AttachVolume{}
+	return a.ForEachVolume(func(volume *AttachVolume) bool {
+		a.volumes[volume.Name] = volume
+		return true
+	})
```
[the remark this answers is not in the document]
lukas016: i reverted these changes

pkg/controllers/machine_controller_volumes.go
```diff
@@ -83,16 +84,17 @@ func getLastVolumeSize(machine *api.Machine, volumeID string) *int64 {
...
 	specVolumes := r.listDesiredVolumes(machine)
 
-	currentVolumeNames := sets.New[string]()
+	currentVolumeNames := sets.NewString()
 	if err := attacher.ForEachVolume(func(volume *AttachVolume) bool {
 		currentVolumeNames.Insert(volume.Name)
 		return true
 	}); err != nil {
-		return nil, fmt.Errorf("error iterating attached volumes: %w", err)
+		return nil, fmt.Errorf("error iteratin attached volumes: %w", err)
```
lukas016, line 97: type

lukas016:
problems solve in this PR will solve with events and different logic.

PR #146, opened 2024-01-31
Graceful shutdown

lukasfrank:
# Proposed Changes

-
-
-

Fixes #

### DIFF
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -360,17 +360,16 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger, 
 	}
 
 	if machine.Spec.ShutdownAt.IsZero() {
-		return r.shutdownMachine(ctx, log, machine, domain)
+		machine.Spec.ShutdownAt = time.Now()
+		if _, err := r.machines.Update(ctx, machine); err != nil {
+			return false, fmt.Errorf("failed to update ShutdownAt: %w", err)
+		}
+		log.V(1).Info("Updated machine ShutdownAt", "ShutdownAt", machine.Spec.ShutdownAt)
 	}
 
 	if machine.Spec.ShutdownAt.Add(r.gcVMGracefulShutdownTimeout).Before(time.Now()) {
-		state, _, _, _, _, err := r.libvirt.DomainGetInfo(domain)
-		if libvirt.IsNotFound(err) {
-			return false, nil
-		}
-
-		log.V(1).Info("Wait to complete shutdown", "State", state, "ShutdownAt", machine.Spec.ShutdownAt)
-		return true, nil
+		//Trigger machine shutdown until VMGracefulShutdownTimeout is over
+		return r.shutdownMachine(log, machine, domain)
 	}
 
 	return false, r.destroyDomain(log, domain)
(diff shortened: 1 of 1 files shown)


pkg/controllers/machine_controller.go
```diff
@@ -312,92 +312,97 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
...
+	domain := libvirt.Domain{
+		UUID: libvirtutils.UUIDStringToBytes(machine.ID),
+	}
 
 	if machine.Spec.ShutdownAt.IsZero() {
-		machine.Spec.ShutdownAt = time.Now()
-		if _, err := r.machines.Update(ctx, machine); err != nil {
-			return fmt.Errorf("failed to set ShutdownAt: %w", err)
-		}
+		return r.shutdownMachine(ctx, log, machine, domain)
```
lukas016, line 363: call shutdown multiple times can be useful what i heard from @gehoern because sometimes signal can be ignore.

pkg/controllers/machine_controller.go
```diff
@@ -312,92 +312,97 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
...
-	log.V(1).Info("Starting additional resources cleanup")
-
-	if err := r.deleteVolumes(ctx, log, machine); err != nil {
-		return fmt.Errorf("failed to remove machine disks: %w", err)
+func (r *MachineReconciler) destroyDomain(log logr.Logger, domain libvirt.Domain) error {
+	if err := r.libvirt.DomainDestroy(domain); err != nil {
+		if libvirt.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to initiate foreful shutdown: %w", err)
```
lukas016, line 384: typo: forceful

pkg/controllers/machine_controller.go
```diff
@@ -312,92 +312,97 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
...
 	}
-	log.V(1).Info("Removed machine disks")
 
-	if err := r.deleteNetworkInterfaces(ctx, log, machine); err != nil {
-		return fmt.Errorf("failed to remove machine network interfaces: %w", err)
-	}
-	log.V(1).Info("Removed network interfaces")
+	log.V(1).Info("Triggered forceful shutdown")
```
lukas016, line 387: Are you sure with Triggered? It looks as destroy is still running but destroy is blocking call and it has to be finished on this line.

pkg/controllers/machine_controller.go
```diff
@@ -312,92 +312,97 @@ func (r *MachineReconciler) startGarbageCollector(ctx context.Context, log logr.
...
+	return false, r.destroyDomain(log, domain)
 }
 
-func (r *MachineReconciler) cleanupAdditionalResources(ctx context.Context, log logr.Logger, machine *api.Machine) error {
-	log.V(1).Info("Starting additional resources cleanup")
-
-	if err := r.deleteVolumes(ctx, log, machine); err != nil {
-		return fmt.Errorf("failed to remove machine disks: %w", err)
+func (r *MachineReconciler) destroyDomain(log logr.Logger, domain libvirt.Domain) error {
+	if err := r.libvirt.DomainDestroy(domain); err != nil {
```
lukas016, line 380: pls use graceful Destroy (with flags) benefits is flushing of disk cache before delete machine, so you have small change for FS corruption

lukas016 (changes requested):

pkg/controllers/machine_controller.go
lukasfrank: Will be addressed in the main GC PR

PR #150, opened 2024-02-02
Improved `Makefile`

### LINKED ISSUE #91: update makefile according to kubebuilder
# Summary

We use makefile based on the kubebuilder project: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go

Kubebuilder makefile is standard for all our projects in ironcore-dev and the makefile template was updated.
So we need to update our makefile in the project according to the kubebuilder template.

Ideally, I want to keep our installation way of golang tools which is more standardized and it solves the synchronization problem of tools.

## Target

- update makefile
- fix small issues (if it is needed)

## Motivation

- more standardized project

lukas016:
# Proposed Changes

- Unified  `Makefile` with the kubebuilder `Makefile`
- Bugfix in `Makefile`
- Create development deployment
- Generalize default deployment

Fixes #91

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,9 +10,17 @@ MKDOCS_IMG=onmetal/libvirt-provider-docs
 # Code depend on OS
 TARGET_OS ?= linux
 TARGET_ARCH ?= amd64
+CGO_ENABLED ?= 1
 
-LIBVIRTPROVIDERBIN=$(LOCALBIN)/libvirt-provider
-LIBVIRTPROVIDERMAINPATH=$(shell pwd)/provider/cmd
+LIBVIRT_PROVIDER_BIN=$(LOCALBIN)/libvirt-provider
+LIBVIRT_PROVIDER_BIN_SOURCE=./provider/cmd
+
+GITHUB_PAT_PATH ?=
+ifeq (,$(GITHUB_PAT_PATH))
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -77,92 +84,108 @@ check-license: addlicense ## Check that every file has a license header present.
...
+
+ifndef ignore-not-found
+  ignore-not-found = false
+endif
+
 .PHONY: deploy
 deploy: kustomize ## Deploy libvirt-provider into the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	kubectl apply -k config/default
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
```
lukasfrank, line 151: Is there a reason for chaining the commands?

lukas016: i followed kubebuilder approach here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L223

lukasfrank (approved):

so-sahu (approved):

afritzler (approved):

PR #163, opened 2024-02-08
Handling Empty Files Before Unmarshalling to Prevent Unmarshal Errors

### LINKED ISSUE #155: Handle Empty Files Before Unmarshalling to Prevent Unmarshal Errors
**Describe the bug**
We have several features implemented, such as `startCheckAndEnqueueVolumeResize`, `startCheckAndEnqueueMachineState`, and `startGarbageCollector`, that operate at regular intervals in separate goroutines. These features involve listing machines in the store. However, during the investigation of the inconsistent behavior observed in integration tests, it has been observed that although it is rare, the error `failed to list machines {"error": "failed to read object: failed to unmarshal object: unexpected end of JSON input"}` occurs occasionally.

**Root Cause Analysis**
Upon investigation, it was discovered that this error primarily occurs because the code retrieves a file from the store and attempts to unmarshal it directly without first checking if the file is empty. When the code tries to unmarshal an empty file, the aforementioned error occurs.

**Proposed Solution**
The proposed solution is to implement a check for an empty file before unmarshalling it. If the file is empty, the system should return an empty object instead of attempting to unmarshal it.

**Expected Outcome**
With the implementation of the proposed solution, we anticipate the elimination of the unexpected end of JSON input error. This will enhance the stability and reliability of the system, ensuring consistent behavior during integration tests and regular operations.

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/155

### DIFF
--- a/pkg/host/store.go
+++ b/pkg/host/store.go
@@ -68,7 +68,7 @@ type CreateStrategy[E api.Object] interface {
 	PrepareForCreate(obj E)
 }
 
-func (s *Store[E]) Create(ctx context.Context, obj E) (E, error) {
+func (s *Store[E]) Create(_ context.Context, obj E) (E, error) {
 	s.idMu.Lock(obj.GetID())
 	defer s.idMu.Unlock(obj.GetID())
 
@@ -100,7 +100,10 @@ func (s *Store[E]) Create(ctx context.Context, obj E) (E, error) {
 	return obj, nil
 }
 
-func (s *Store[E]) Get(ctx context.Context, id string) (E, error) {
+func (s *Store[E]) Get(_ context.Context, id string) (E, error) {
+	s.idMu.Lock(id)
+	defer s.idMu.Unlock(id)
+
 	object, err := s.get(id)
 	if err != nil {
 		return utils.Zero[E](), fmt.Errorf("failed to read object: %w", err)
@@ -109,7 +112,7 @@ func (s *Store[E]) Get(ctx context.Context, id string) (E, error) {
 	return object, nil
 }
 
-func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
+func (s *Store[E]) Update(_ context.Context, obj E) (E, error) {
 	s.idMu.Lock(obj.GetID())
 	defer s.idMu.Unlock(obj.GetID())
 
@@ -143,7 +146,7 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 	return obj, nil
 }
 
-func (s *Store[E]) Delete(ctx context.Context, id string) error {
+func (s *Store[E]) Delete(_ context.Context, id string) error {
 	s.idMu.Lock(id)
 	defer s.idMu.Unlock(id)
 
@@ -183,7 +186,7 @@ func (s *Store[E]) List(ctx context.Context) ([]E, error) {
 			continue
 		}
 
-		object, err := s.get(entry.Name())
+		object, err := s.Get(ctx, entry.Name())
 		if err != nil {
 			return nil, fmt.Errorf("failed to read object: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


pkg/host/store.go
```diff
@@ -221,6 +221,10 @@ func (s *Store[E]) get(id string) (E, error) {
 	}
 
 	obj := s.newFunc()
+	if len(file) == 0 {
```
lukas016, line 224: i prefer maybe add mutex.Lock for getting of file content

lukasfrank: I agree, please use a `RWMutex` in the store.

pkg/host/store.go
```diff
@@ -221,6 +221,10 @@ func (s *Store[E]) get(id string) (E, error) {
...
+		return obj, err
+	}
+
 	if err := json.Unmarshal(file, &obj); err != nil {
 		return utils.Zero[E](), fmt.Errorf("failed to unmarshal object: %w", err)
```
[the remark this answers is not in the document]
so-sahu: Added

pkg/host/store.go
so-sahu: Added.

pkg/host/store.go
```diff
@@ -211,6 +211,9 @@ func (s *Store[E]) Watch(ctx context.Context) (store.Watch[E], error) {
 }
 
 func (s *Store[E]) get(id string) (E, error) {
+	s.idMu.RLock(id)
```
lukas016, line 214: it cannot work. You will create deadlock in update function. In update function you locked mutex for writting so RLock will wait here for unlock of write operation which cannot be done because it will be block RLock function.

pkg/sync/sync.go
```diff
@@ -25,41 +25,69 @@ func NewMutexMap[K comparable]() *MutexMap[K] {
...
-
 	entry.mu.Lock()
+	for entry.readerCnt > 0 {
+		entry.readerDone.Wait()
+	}
 }
 
-// Unlock unlocks the given key.
+// Unlock unlocks the given key for writing.
 func (m *MutexMap[K]) Unlock(key K) {
```
lukas016, line 47: why we cannot use existing solution?

lukas016 (changes requested):

pkg/host/store.go
so-sahu: No, it won't lead to deadlock. as get will increase the redercount and write will wait for it to complete and then write. I validated it. It's working fine.

It would have led to deadlock in current implementations due to the reason you mentioned.

pkg/sync/sync.go
so-sahu: If we try to use existing solution here, it would lead to deadlock due to the same reason you mentioned. Hence, I have used conditional synchronization..

pkg/host/store.go
lukas016: but you can create starvation problem in this case, i think. if you will get very quickly new reading operation, write can be postponed forever.

so-sahu: In theory, it's possible. However, I haven't encountered such an issue even after running multiple instances of consecutive integration tests, especially considering the already super-fast read operations. Therefore, I believe a client sending GET requests faster than the this might not be a practical scenario.

However, we can always refactor if we see any such situation in real-time in future.

so-sahu: As discussed offline, there's no need to over-engineer in this case. Therefore, proceeding with a simple solution in the latest commit.

pkg/sync/sync.go
so-sahu: Using the existing solution now.

lukas016 (approved):
/LGTM

hardikdr (approved):
Thanks for the PR !! 
Let's keep an eye on how the locks during reads play out. If things get slow, we might want to adjust later.

PR #169, opened 2024-02-09
Added `Terminating` & `Terminated` machine state 

### LINKED ISSUE #158: Implement `Terminating` machine state
# Summary

The `ironcore` project supports now a `Terminating` state. The `libvirt-provider` should also provide this information.

/ref https://github.com/ironcore-dev/ironcore/pull/984

lukas016:
# Proposed Changes

- Implement `terminating `state
- Implement `terminated` state
- Report `blocked` state as `pending`

Fixes #158


pkg/controllers/machine_controller.go
```diff
@@ -329,6 +330,11 @@ func (r *MachineReconciler) processMachineDeletion(ctx context.Context, log logr
 		return fmt.Errorf("failed to delete machine: %w", err)
 	}
 	log.V(1).Info("Deleted machine")
+	machine.Status.State = api.MachineStateTerminated
```
lukasfrank, line 333: What about moving `machine.Status.State = api.MachineStateTerminated` down to the removal of the finalizers?

lukas016: if you think line 353, i don't think it is correct. Terminated state mean machine was deleted and cannot be start again. If you will move it down, machine will be in state terminating and it can be stuck on error during removing of leftovers (volumes, nics, ...) and you will report to customer state terminating which mean machine is in shutdown process what isn't true.

lukasfrank (dismissed):

afritzler:
Can we have a more meaningful PR title like: Add `Terminating` state handling

lukas016:
> Can we have a more meaningful PR title like: Add `Terminating` state handling

yes, but technically i added more states

pkg/controllers/machine_controller.go
```diff
@@ -360,11 +366,12 @@ func (r *MachineReconciler) deleteMachine(ctx context.Context, log logr.Logger,
...
 
 	if machine.Spec.ShutdownAt.IsZero() {
+		machine.Status.State = api.MachineStateTerminating
 		machine.Spec.ShutdownAt = time.Now()
 		if _, err := r.machines.Update(ctx, machine); err != nil {
-			return false, fmt.Errorf("failed to update ShutdownAt: %w", err)
+			return false, fmt.Errorf("failed to update ShutdownAt and State: %w", err)
 		}
-		log.V(1).Info("Updated machine ShutdownAt", "ShutdownAt", machine.Spec.ShutdownAt)
+		log.V(1).Info("Updated machine state", "ShutdownAt", machine.Spec.ShutdownAt, "State", machine.Status.State)
```
so-sahu, line 374: Can we replace `state` with `object` here, as ShutdownAt is part of the spec, and state is a part of the status.

lukas016: What do you think about: Updated API machine?
I think we have similar strings somewhere else in code.

so-sahu: That sounds good to me.

so-sahu (approved):
LGTM

PR #176, opened 2024-02-12
Implemented qemu guest agent

### LINKED ISSUE #165: qemu guest agent support
# Summary

We need to implement support for [qemu guest agent](https://wiki.libvirt.org/Qemu_guest_agent.html)
This feature has to be optional and it can be enabled/disabled with argument.
We can inspire by virtlet implementation: https://github.com/onmetal/virtlet/blob/main/scheduler/scheduler.go#L124

## Basic example

If feature is enabled, we can add into specification path to socket QemuGuestAgentSocket (type string) or use type bool. I prefer use empty string for disable feature and path to socket for enable feature (it can be useful for ops guys).
I recommend use machine folder as default storage for socket.

## Motivation

We require this feature for manage/monitoring VM and getting additional information from VM.

lukas016:
# Proposed Changes

- Extend `api.Machine` spec and status
- Add new argument
- Add new logic into server and reconciler

Fixes #165


pkg/api/machine.go
```diff
@@ -28,13 +28,16 @@ type MachineSpec struct {
 	NetworkInterfaces []*NetworkInterfaceSpec `json:"networkInterfaces"`
 
 	ShutdownAt time.Time `json:"shutdownAt,omitempty"`
+
+	QemuGuestAgentEnable bool
```
lukasfrank, line 32:
```suggestion
	QemuGuestAgent GuestAgent
```

Can we maybe update the spec? 
```
type GuestAgent string 
const (
	EnableGuestAgent GuestAgent = "Enable"
	DisableGuestAgent GuestAgent = "Disable"
)
```

pkg/api/machine.go
```diff
@@ -28,13 +28,16 @@ type MachineSpec struct {
...
 }
 
 type MachineStatus struct {
 	VolumeStatus           []VolumeStatus           `json:"volumeStatus"`
 	NetworkInterfaceStatus []NetworkInterfaceStatus `json:"networkInterfaceStatus"`
 	State                  MachineState             `json:"state"`
 	ImageRef               string                   `json:"imageRef"`
+	QemuGuestAgent         *AgentStatus             `json:"qemuGuestAgent,omitempty"`
```
lukasfrank, line 40: In general: Do we need `Qemu` in `QemuGuestAgent` ? Isn't that implementation detail?

lukas016: Theoretically you can have multiple type of agents (https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.3/html/virtual_machine_management_guide/installing_guest_agents_and_drivers_linux#doc-wrapper). But we support qemu currently so i wanted do it extendable for future.

pkg/api/machine.go
lukas016: What are benefits? Because we have some approach for Hugepages. Do will we need more values as true/false? Compare of strings can has higher impact to performance too.

pkg/api/machine.go
lukas016: but we can do it with extend agentStatus, so yes we can rename it to GuestAgent.

pkg/api/machine.go
lukasfrank: These are the k8s api conventions. The general reasoning is that the string types can be extended if needed.

pkg/api/machine.go
lukasfrank: I think it's a good idea to put the type of the `GuestAgent` into the status.

pkg/api/machine.go
lukas016: we can go this way, but i prefer lower case letters. It is simpler for comparison and i don't have to do conversation or support multiple variant.

lukas016: i  will implement it as you suggest for keep the same format. Conversion to lower case we can discuss next time.

pkg/api/machine.go
lukas016: fixed

pkg/api/machine.go
lukas016: fixed

provider/server/server.go
```diff
@@ -68,9 +72,24 @@ func setOptionsDefaults(o *Options) {
...
+}
+
 func New(opts Options) (*Server, error) {
 	setOptionsDefaults(&opts)
 
+	guestAgent, err := convertGuestAgentFromString(opts.GuestAgent)
+	if err != nil {
+		return nil, err
+	}
+
```
lukasfrank, line 92: wdyt about passing the `GuestAgent` here and to the validation outside of the server?

pkg/api/machine.go
```diff
@@ -107,3 +121,8 @@ const (
 	NetworkInterfaceStatePending  NetworkInterfaceState = "Pending"
 	NetworkInterfaceStateAttached NetworkInterfaceState = "Attached"
 )
+
+type AgentStatus struct {
+	Type GuestAgent `json:"type"`
```
lukasfrank, line 126: Does this bring us any benefit having this here again? If no: I prefer to have the `AgentStatus` as a pointer in the `MachineStatus`

lukasfrank (changes requested):

provider/server/server.go
```diff
@@ -68,9 +72,24 @@ func setOptionsDefaults(o *Options) {
 	}
 }
 
+func convertGuestAgentFromString(guestAgent string) (api.GuestAgent, error) {
+	availableGuestAgents := api.GuestAgentAvailable()
```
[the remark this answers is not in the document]
lukas016: i had switch as first solution but you have to handle options in 2 packages (api and here). I don't like it. It can be harder handle it in future and this function is calling just one time during start, so i don't see big benefits in performance improvement with switch.

pkg/api/machine.go
lukas016: No significant benefits, we can say you don't need handle nil for GuestAgent in status in future. I can revert it.

provider/server/server.go
lukas016: GuestAgent is currently used in server only. If you want to implement conversion somewhere else, it can be do, but we can follow my suggestion with implement pflag.Value interface because in other case we will end with multiple convert functions this way and implementation of pflag.Value is fast failure.

lukas016: if you don't like use API type for that. We can move GuestAgent into new package. Maybe something as features/guestagent ? And it can be import into provider/cmd/app and pkg/api as independent package and we can without big problems implement pflag.Value interface

provider/server/machine_create.go
```diff
@@ -89,6 +89,8 @@ func (s *Server) createMachineFromIRIMachine(ctx context.Context, log logr.Logge
 		machine.Spec.Image = &iriMachine.Spec.Image.Image
 	}
 
+	machine.Spec.GuestAgent = s.guestAgent
```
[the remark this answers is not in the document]
lukas016: fixed

pkg/api/machine.go
lukas016: fixed

provider/server/server.go
lukas016: reimplemented as pflag.Value interface:

```
Error: invalid argument "Test" for "--guest-agent-type" flag: unsupported option Test
Usage:
  libvirt-provider [flags]

Flags:
      --address string                             Address to listen on. (default "/var/run/iri-machinebroker.sock")
      --apinet-kubeconfig string                   Path to the kubeconfig file for the apinet-cluster.
      --apinet-node-name string                    APInet node name
```

provider/server/server.go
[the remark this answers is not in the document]
lukas016: GuestAgentAvailable was moved

provider/server/server.go
lukas016: fixed

provider/cmd/app/option_types.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (g *GuestAgentOption) GetAPIGuestAgent() api.GuestAgent {
+	if g == nil || *g == "" {
+		return api.GuestAgentNone
+	}
+	return api.GuestAgent(*g)
+}
+
+func GuestAgentOptionAvailable() []string {
```
so-sahu, line 46: Scope of this function can be reduced to package level, as it's only being used within this package.

pkg/api/machine.go
```diff
@@ -28,13 +28,23 @@ type MachineSpec struct {
...
+	GuestAgentNone GuestAgent = "None"
+	GuestAgentQemu GuestAgent = "Qemu"
+)
+
 type MachineStatus struct {
 	VolumeStatus           []VolumeStatus           `json:"volumeStatus"`
 	NetworkInterfaceStatus []NetworkInterfaceStatus `json:"networkInterfaceStatus"`
 	State                  MachineState             `json:"state"`
 	ImageRef               string                   `json:"imageRef"`
+	GuestAgent             *GuestAgentStatus        `json:"guestAgent,omitempty"`
```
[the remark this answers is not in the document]
lukas016: fixed

provider/cmd/app/option_types.go
lukas016: fixed

so-sahu (dismissed):
LGTM

pkg/controllers/machine_controller.go
```diff
@@ -806,6 +811,38 @@ func (r *MachineReconciler) setTCMallocPath(domain *libvirtxml.Domain) error {
 	return nil
 }
 
+func (r *MachineReconciler) setGuestAgent(machine *api.Machine, domainDesc *libvirtxml.Domain) {
+	if machine.Spec.GuestAgent == api.GuestAgentNone {
```
hardikdr, line 815: Can we have this change outside, so that we know code doesn't enter at all if it's disabled? Just a cosmetic suggestion.

hardikdr (commented):
Thanks @lukas016 for the PR, just a minor suggestion, looks good otherwise.

pkg/controllers/machine_controller.go
lukas016: how do you mean that? I prefer have all logic in function. It is logical block, so you have to just delete this function and calls to it and you don't need study code what else you have to delete.

hardikdr: I would usually prefer to keep the optional features behind the `if`, then we know code-flow never enters into this method when flag is set, and main flow is safe-guarded. 
In this case, it's an optional suggestion, feel free to resolve it otherwise. 
```golang
if conditionMet {
   setGuestAgent()
}
```

hardikdr (dismissed):

pkg/api/machine.go
lukasfrank:
```suggestion
```

pkg/controllers/machine_controller.go
```diff
@@ -790,6 +797,34 @@ func (r *MachineReconciler) setTCMallocPath(domain *libvirtxml.Domain) error {
...
+		},
+		Target: &libvirtxml.DomainChannelTarget{
+			VirtIO: &libvirtxml.DomainChannelTargetVirtIO{
+				Name: "org.qemu.guest_agent.0",
+			},
+		},
+	}
+
+	domainDesc.Devices.Channels = append(domainDesc.Devices.Channels, agent)
+	machine.Status.GuestAgentStatus = &api.GuestAgentStatus{Type: machine.Spec.GuestAgent, Addr: "unix://" + socketPath}
```
lukasfrank, line 825: We do not use this property 
```suggestion
	machine.Status.GuestAgentStatus = &api.GuestAgentStatus{ Addr: "unix://" + socketPath}
```

pkg/controllers/machine_controller.go
```diff
@@ -888,3 +923,11 @@ func machineDomain(machineID string) libvirt.Domain {
 		UUID: libvirtutils.UUIDStringToBytes(machineID),
 	}
 }
+
+func getShutdownMode(agent *api.GuestAgentStatus) libvirt.DomainShutdownFlagValues {
```
lukasfrank, line 927: What do you think about inlining this few lines?

lukasfrank (changes requested):

pkg/controllers/machine_controller.go
lukas016: it is information for user. Do you really want to delete it?

lukas016: wait, i used status.Type in getShutdownMode

pkg/controllers/machine_controller.go
lukas016: theoretically, i prefer more function for simpler extending but it this case no problem.
If you want to simplify it, we can use automode and libvirt will choose best option alone, but i am afraid of unexpected behavior in this case and logic will depend on another component and behavior can change base on version what can bring more problem for us. So i don't recommend use automode.

pkg/controllers/machine_controller.go
```diff
@@ -678,6 +681,10 @@ func (r *MachineReconciler) domainFor(
 		return nil, nil, nil, err
 	}
 
+	if machine.Spec.GuestAgent != api.GuestAgentNone {
```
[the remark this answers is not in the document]
lukas016: ok logic is stable because it is set in call CreateMachine.

pkg/controllers/machine_controller.go

pkg/controllers/machine_controller.go

lukasfrank (approved):

PR #189, opened 2024-02-19
Honor graceful shutdown in `deleteMachine` integration test

### LINKED ISSUE #175: improve integration test for deleteMachine
# Summary

We need to update delete machine test: https://github.com/ironcore-dev/libvirt-provider/blob/main/provider/server/machine_delete_test.go#L17)

We will need to cover these 3 scenarios:
- Destroy machine without graceful shutdown
- Graceful shutdown without destroy
- Destroy machine with graceful shutdown

We implemented new states Terminating and Terminated. Terminating state has to be use in scenarios:
- Graceful shutdown without destroy
- Destroy Machine with graceful shutdown

Terminated state can be little worse catch because it is set after delete machine and before starting deletion of leftovers (volumes, nics, ...)

Currently we support graceful shutdown and we have implemented scenario: Destroy machine with graceful shutdown

## Motivation

Better cover of use cases in integration test

Rohit-0505:
- add test validation for machine terminating state
- add test validation for vm graceful shutdown 
- add test validation for vm destroy after graceful shutdown period

Fixes #175 (partially)


provider/server/machine_delete_test.go
```diff
@@ -17,8 +18,122 @@ import (
...
+		createResp, err := machineClient.CreateMachine(ctx, &iri.CreateMachineRequest{
+			Machine: &iri.Machine{
+				Metadata: &irimeta.ObjectMetadata{
+					Labels: map[string]string{
+						"foo": "bar",
+					},
+				},
+				Spec: &iri.MachineSpec{
+					Power: iri.Power_POWER_ON,
+					Image: &iri.ImageSpec{
```
lukas016, line 32: you have to use ceph volume with os. This option cannot be use in this case.

provider/server/server_suite_test.go
```diff
@@ -28,13 +28,14 @@ import (
...
-	eventuallyTimeout    = 30 * time.Second
-	pollingInterval      = 50 * time.Millisecond
-	consistentlyDuration = 1 * time.Second
-	machineClassx3xlarge = "x3-xlarge"
-	machineClassx2medium = "x2-medium"
-	baseURL              = "http://localhost:8080"
-	streamingAddress     = "127.0.0.1:20251"
+	eventuallyTimeout       = 30 * time.Second
+	pollingInterval         = 50 * time.Millisecond
+	gracefulShutdownTimeout = 20 * time.Second
```
lukas016, line 33: it is too small for graceful shutdown

provider/server/machine_delete_test.go
```diff
@@ -80,6 +195,18 @@ var _ = Describe("DeleteMachine", func() {
...
+		Eventually(func() iri.MachineState {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(listResp.Machines).Should(HaveLen(1))
+			return listResp.Machines[0].Status.State
+		}).WithTimeout(gracefulShutdownTimeout).WithPolling(1 * time.Second).Should(Equal(iri.MachineState_MACHINE_TERMINATING))
```
lukas016, line 208: i don't think, this is correct logic. You check state in loop and you end, if state is terminating, but state will be terminating immediately after send shutdown command. I recommend use ShutdownAt field in spec. ShutdownAt + gracefulShutdownTimeout is graceful shutdown period.

provider/server/machine_delete_test.go
Rohit-0505: I was just trying to get the vm booted with `ceph-volume` and OSimage for testing purpose, However we cannot do the ceph-volume boot at the moment because it would be required  to setup `ceph-provider` as well.
I will anyway proceed with ram boot for the integration test.

provider/server/server_suite_test.go
Rohit-0505: We have given `eventuallyTimeout` as 30s, which seems to be sufficient. So we have to give the `gracefulShutdownTimeout` anything less than 30s here.

provider/server/machine_delete_test.go
lukas016: do you have image which is bootable with image in spec?

provider/server/server_suite_test.go
lukas016: it depends on os Image, i recommend create new app run for deletion tests.

provider/server/server_suite_test.go
```diff
@@ -28,13 +28,16 @@ import (
...
-	machineClassx2medium = "x2-medium"
-	baseURL              = "http://localhost:8080"
-	streamingAddress     = "127.0.0.1:20251"
+	eventuallyTimeout       = 80 * time.Second
+	pollingInterval         = 50 * time.Millisecond
+	gracefulShutdownTimeout = 60 * time.Second
+	consistentlyDuration    = 1 * time.Second
+	machineClassx3xlarge    = "x3-xlarge"
+	machineClassx2medium    = "x2-medium"
+	squashfsOsImage         = "ghcr.io/ironcore-dev/ironcore-image/gardenlinux:squashfs-dev-20240123-v2"
```
lukas016, line 37: acronyms in golang are written with uppercase: squashfsOSImage
More info here: https://google.github.io/styleguide/go/decisions.html#initialisms

provider/server/machine_delete_test.go
```diff
@@ -14,11 +15,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// TODO: This test will require update after merge of PR: #101
 var _ = Describe("DeleteMachine", func() {
 
-	It("should delete a machine", func(ctx SpecContext) {
-		By("creating a machine")
+	It("should delete a machine with graceful shutdown", func(ctx SpecContext) {
+		By("creating a machine using squashfs osImage which boots properly ")
```
lukas016, line 21:
```suggestion
		By("creating a machine using squashfs os image which boots properly ")
```

provider/server/machine_delete_test.go
```diff
@@ -28,12 +28,116 @@ var _ = Describe("DeleteMachine", func() {
...
+		By("ensuring machine is gracefully shutdown")
+		Eventually(func() int {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			return len(listResp.Machines)
+		}).Within(gracefulShutdownTimeout).ProbeEvery(10).Should(BeZero())
```
lukas016, line 108: it doesn't work properly. you cannot be sure, if machine was deleted gracefully or destroy.

provider/server/machine_delete_test.go
```diff
@@ -28,12 +28,116 @@ var _ = Describe("DeleteMachine", func() {
...
+		createResp, err := machineClient.CreateMachine(ctx, &iri.CreateMachineRequest{
+			Machine: &iri.Machine{
+				Metadata: &irimeta.ObjectMetadata{
+					Labels: map[string]string{
+						"foo": "bar",
+					},
+				},
+				Spec: &iri.MachineSpec{
+					Power: iri.Power_POWER_ON,
+					Image: &iri.ImageSpec{
```
lukas016, line 132: pls remove image completely, it won't be booted on 100%

provider/server/machine_delete_test.go
```diff
@@ -80,7 +184,30 @@ var _ = Describe("DeleteMachine", func() {
...
+		By("ensuring machine is not deleted until gracefulShutdownTimeout ")
+		Eventually(func() []*iri.Machine {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			return listResp.Machines
+		}).Within(gracefulShutdownTimeout).ProbeEvery(5).Should(HaveLen(1))
```
lukas016, line 208: i don't think this condition is correct. During graceful period you will get machineList with one machine, so eventually will do only 1 loop.

lukas016 (changes requested):
logic isn't correct. I recommend create for delete machine start own libvirt-provider with different arguments. it will be for simulate every cases.

provider/server/machine_delete_test.go
```diff
@@ -14,11 +15,111 @@ import (
...
+		By("ensuring machine is gracefully shutdown")
+		Eventually(func() int {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			return len(listResp.Machines)
+		}).Within(gracefulShutdownTimeout).ProbeEvery(2).Should(BeZero())
```
lukas016, line 108: could you pls write here comment something as: ProbeEvery has to be ideally less or equal of half ResyncIntervalGarbageCollector

lukas016 (dismissed):
LGTM

provider/server/machine_delete_test.go
```diff
@@ -14,11 +15,111 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// TODO: This test will require update after merge of PR: #101
 var _ = Describe("DeleteMachine", func() {
 
-	It("should delete a machine", func(ctx SpecContext) {
-		By("creating a machine")
+	It("should delete a machine with graceful shutdown", func(ctx SpecContext) {
+		By("creating a machine using squashfs os image which boots properly")
```
afritzler, line 21:
```suggestion
		By("creating a machine using squashfs os image")
```

provider/server/machine_delete_test.go
```diff
@@ -14,11 +15,111 @@ import (
...
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(listResp.Machines).NotTo(BeEmpty())
+			return listResp.Machines[0].Status.State
+		}).Should(Equal(iri.MachineState_MACHINE_RUNNING))
+
+		//allow some time for the vm to boot properly
+		time.Sleep(30 * time.Second)
```
afritzler, line 80: Sleep is good enough for now but is not the best solution here. Alternatively we could think of using `Eventually` and try to wait until the SSH port of the machine is open. We should be able to use a local network device as a `NetworkInterface` here.

provider/server/machine_delete_test.go
```diff
@@ -14,11 +15,111 @@ import (
...
+		By("ensuring machine is gracefully shutdown")
+		Eventually(func() int {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			return len(listResp.Machines)
+		}).Within(gracefulShutdownTimeout).ProbeEvery(2 * time.Second).Should(BeZero()) // ProbeEvery has to be ideally less than or equal to half of ResyncIntervalGarbageCollector
```
afritzler, line 108: Can we use a constant here in the same way we are using it for the graceful shutdown timeout?

afritzler (dismissed):

provider/server/machine_delete_test.go
lukas016: we can, but different tests have different Probe time, Only this one has to be shorter for now.

provider/server/machine_delete_test.go
lukas016: @Rohit-0505 could you pls create issue for it. we will have the same problem in another tests too. So we need one proper solution for boot waiting.

Rohit-0505: yeah I will create one

provider/server/machine_delete_test.go
```diff
@@ -80,7 +181,29 @@ var _ = Describe("DeleteMachine", func() {
...
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(listResp.Machines).Should(HaveLen(1))
+			return listResp.Machines[0].Status.State
+		}).Should(Equal(iri.MachineState_MACHINE_TERMINATING))
+
+		By("ensuring machine is not deleted until gracefulShutdownTimeout")
+		time.Sleep(gracefulShutdownTimeout)
```
so-sahu, line 197: Here, instead of using time.Sleep() to wait for the machine not to be deleted until gracefulShutdownTimeout, we can improve the test by checking that the machine exists consistently over that interval. This approach ensures that we continuously verify the machine's existence instead of waiting passively.

Rohit-0505: @so-sahu I had an offline discussion with @lukas016 on this and decided to go ahead with time.sleep() for now.

lukas016: it doesn't make sense check it in loop, machine has to still exist after graceful period in this test.

lukas016:
@Rohit-0505 i didn't see any comment for first use case from issue here. Pls don't forget create follow up issue for it

Rohit-0505:
> @Rohit-0505 i didn't see any comment for first use case from issue here. Pls don't forget create follow up issue for it

I discussed that with @afritzler , If we really want to cover the scenario of direct destroy machine by starting a new libvirt app run with `gracefulShutdownTimeout=0` other than one in `server_suite_test` already present. He had different opinion about this and wanted to clarify over here.
@afritzler can you please give your thoughts here?

provider/server/machine_delete_test.go
so-sahu: As discussed with @lukas016 offline, the sleep currently ensures that the machine is not deleted until the gracefulShutdownTimeout. Instead of passively assuming that the machine still exists, it would be beneficial to actively check for its existence during the graceful timeout period using a `consistently` block.

provider/server/machine_delete_test.go
```diff
@@ -80,7 +181,31 @@ var _ = Describe("DeleteMachine", func() {
...
+		Consistently(func() bool {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(listResp.Machines).Should(HaveLen(1))
+			return err == nil && len(listResp.Machines) == 1
+		}, gracefulShutdownTimeout, consistentlyDuration).Should(BeTrue())
```
lukas016, line 206: Pls use methods for set consistently behavior: https://onsi.github.io/gomega/#consistently

consistentlyDuration is wrong name because it isn't duration but interval for polling.

provider/server/machine_delete_test.go
```diff
@@ -80,7 +181,31 @@ var _ = Describe("DeleteMachine", func() {
...
+		}).Should(Equal(iri.MachineState_MACHINE_TERMINATING))
+
+		By("ensuring machine is not deleted until gracefulShutdownTimeout")
+		Consistently(func() bool {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
```
lukas016, line 203: Fix every Expects in Eventually and Consistently blocks as i presented today

lukas016 (changes requested):

provider/server/machine_delete_test.go
```diff
@@ -80,14 +183,39 @@ var _ = Describe("DeleteMachine", func() {
...
+		By("ensuring machine is not deleted until gracefulShutdownTimeout")
+		Consistently(func(g Gomega) bool {
+			listResp, err := machineClient.ListMachines(ctx, &iri.ListMachinesRequest{
+				Filter: &iri.MachineFilter{
+					Id: createResp.Machine.Metadata.Id,
+				},
+			})
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(listResp.Machines).Should(HaveLen(1))
+			return err == nil && len(listResp.Machines) == 1
```
lukas016, line 207: Are you sure? it looks like combination of 2 way for assertion.

provider/server/machine_delete_test.go
lukas016: this isn't solve

lukas016 (dismissed):
LGTM

lukasfrank (approved):

PR #190, opened 2024-02-19
Replace machine state sync with libvirt lifecycle event subscription

### LINKED ISSUE #177: integrate support for libvirt events
# Summary

Libvirt service supports register event handler, so we can listen to events from libvirt directly in  libvirt-provider and we don't need active pulling of information from libvirt.

              I would suggest to subscribe to the `LifeCycleEvents` of the libvirt Domain and enqueue Machines if something changes in the domain.

Here is some sample code how this might look like:

```
// Subscribe to lifecycle events using the provided context
eventChan, err := l.LifecycleEvents(ctx)
if err != nil {
	fmt.Println("Failed to subscribe to lifecycle events:", err)
	return
}

fmt.Println("Subscribed to lifecycle events. Listening for events...")

// Listen for lifecycle events in a separate goroutine
go func() {
	for event := range eventChan {
		fmt.Printf("Received lifecycle event: %+v\n", event)
                // enqueue Machines here and switch over the events
	}
	fmt.Println("Lifecycle event channel closed.")
}()

// Wait for context to be cancelled (e.g., by receiving a signal)
<-ctx.Done()
fmt.Println("Context cancelled. Shutting down...")
```

This can be also used to consolidate other `go routines` which we are starting during the initialization of the provider.

_Originally posted by @afritzler in https://github.com/ironcore-dev/libvirt-provider/pull/144#pullrequestreview-1877261405_

lukas016:
# Proposed Changes

- delete argument for machine-state-sync
- replace machie-state-sync logic with handling libvirt events

Fixes #177


pkg/controllers/machine_controller.go
```diff
@@ -268,35 +265,29 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
-			state, err := r.getMachineState(id)
-			if err != nil {
-				if libvirt.IsNotFound(err) {
-					log.V(1).Info("Failed to retrieve domain. Requeueing", "machineID", id)
-					r.queue.AddRateLimited(id)
-				} else {
-					log.Error(err, "failed to get machine state", "machineID", id)
-				}
-				continue
+	for {
```
so-sahu, line 276: This can be simplified. 
```suggestion
	for ev := range eventChan {
		log.Info("requeue machine id " + ev.Dom.Name)
		r.queue.AddRateLimited(ev.Dom.Name)
	}

	if err := ctx.Err(); err != nil && err != context.Canceled {
		log.Error(err, "context error")
	}
```

pkg/controllers/machine_controller.go
```diff
@@ -268,35 +265,29 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
 }
 
 func (r *MachineReconciler) startCheckAndEnqueueMachineState(ctx context.Context, log logr.Logger) {
```
so-sahu, line 267: A more suitable name for this function could be `monitorLifecycleEvents` or `watchLifecycleEvents`

so-sahu (changes requested):

pkg/controllers/machine_controller.go
lukas016: i prefer keep the same name convention. maybe `startMonitorLifecycleEvents`

pkg/controllers/machine_controller.go
lukas016: you have to first process all events in channel after you can exit. I don't see big benefits in trigger requeue when main context is close and it cannot be processes anymore.

so-sahu: ok. makes sense.

pkg/controllers/machine_controller.go
afritzler: How about `startEnqueMachineByLibvirtEvent`? Because that is essentially what is going to happen here.

pkg/controllers/machine_controller.go
```diff
@@ -268,35 +265,29 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
-					r.queue.AddRateLimited(id)
-				} else {
-					log.Error(err, "failed to get machine state", "machineID", id)
-				}
-				continue
+	for {
+		select {
+		case ev, ok := <-eventChan:
+			if !ok {
+				log.Error(errors.New("missing lifecycle event"), "lifecycle event channel is closed")
```
afritzler, line 280: I would log the following error here: `Errorf("unexpected channel closure while waiting for data")`

pkg/controllers/machine_controller.go
```diff
@@ -268,35 +265,29 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
...
+				return
 			}
 
-			if state != machine.Status.State {
-				r.queue.AddRateLimited(id)
-			}
+			log.Info("requeue machine id " + ev.Dom.Name)
+			r.queue.AddRateLimited(ev.Dom.Name)
```
afritzler, line 285: Is it a good idea to pass all the events here? I am not sure how many of those events are coming in. Can't we create a `filterLibvirtEvents` helper method to catch only the events which are interesting for the machine lifecycle?

afritzler (changes requested):

pkg/controllers/machine_controller.go
lukas016: it isn't completely true because channel can be close when parent context is closed, what isn't unexpected. but no problem

pkg/controllers/machine_controller.go
lukas016: it is same logic as virtlet has: https://github.com/onmetal/virtlet/blob/main/source/source.go#L104

lukas016: i looked at lifecycle events more deeply and for me every "subevents" look usable: https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainEventType
I will extend log message about event ID, so we can monitor it and we can prepare improvements later.
I would like extend your idea with another events: https://github.com/ironcore-dev/libvirt-provider/issues/192

lukas016: partially fixed

so-sahu (dismissed):
LGTM

hardikdr (commented):
Thanks for the PR @lukas016. 

I'm interested in how many events the Libvirt daemon sends and whether it causes unnecessary reconciliations. I see there's already a parallel issue where we can discuss this further.

lukas016:
@hardikdr i applied all of your suggestions

hardikdr (approved):

PR #191, opened 2024-02-19
Add `ResourceVersion` to internal `Machine` type

### LINKED ISSUE #164: information from API machine can be lost (overwritten)
**Describe the bug**
Our store saves whole API machine object into file. Problem is this object can be save from multiple goroutines (reconcile loop, iri server, gc) which can operate with different version of object at the same time.

**To Reproduce**
For simulation i recommend put time.Sleep(1 * time.Time) into updateDomain function and set long interval (5min) for GC.
1. Create machine
2. Wait for calling updateDomain
3. Call delete machine
4. Check store for metadata.DeleteAt field (it has to be set)
5. Wait for finish reconcile loop which sleeps in updateDomain
6. Check store for metadata.DeleteAt field (it will be missing)

Delete action won't be performed.

**Expected behavior**
Object information cannot be overwritten into older state.

kasabe28:
# Proposed Changes

- Get the latest machine from store before updating machine state, volume states and NIC states. Fixes #164


lukas016 (changes requested):
Root cause wasn't solved

pkg/controllers/machine_controller.go
```diff
@@ -468,6 +468,19 @@ func (r *MachineReconciler) reconcileMachine(ctx context.Context, id string) err
 	}
 	log.V(1).Info("Reconciled domain")
 
+	// get latest machine from store before updating it
+	machine, err = r.machines.Get(ctx, id)
```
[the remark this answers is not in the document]
kasabe28: This solution is resolving given scenario in bug. And if there are more such case we need to handle it differently.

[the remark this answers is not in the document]
lukas016: we discussed possible solution offline

pkg/host/store.go
```diff
@@ -86,6 +86,7 @@ func (s *Store[E]) Create(ctx context.Context, obj E) (E, error) {
 	}
 
 	obj.SetCreatedAt(time.Now())
+	obj.SetGeneration(obj.GetGeneration() + 1)
```
lukas016, line 89: what do you think put here directly 1?

pkg/host/store.go
```diff
@@ -129,6 +130,11 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 		return obj, nil
 	}
 
+	if oldObj.GetGeneration() != obj.GetGeneration() {
+		return utils.Zero[E](), fmt.Errorf("failed to update object: generation is not latest")
+	}
+	obj.SetGeneration(obj.GetGeneration() + 1)
```
lukas016, line 136: optional: i prefer create method IncrementGeneration() or create const for 1 because you call this code on multiple places.

pkg/host/store.go
lukas016, line 89: i looked at official kubernetes api concept and they have written this:

```
generation
integer	A sequence number representing a specific generation of the desired state. Populated by the system. Read-only.
```

Desired state is for us spec structure.
Second problem what i see in future is extend of status with ObservedGeneration logic. we don't have ObservedGeneration in status currently but if we add it, we can end in infinity loop how ObservedGeneration will want to achieve Generation.

 So i prefer here don't use generation but extend metedata with resourceVersion and use it.
@so-sahu, @hardikdr, @afritzler what do you think?

provider/server/machine_create_test.go
```diff
@@ -400,7 +400,7 @@ var _ = Describe("CreateMachine", func() {
 			HaveField("Machine.Spec.NetworkInterfaces", ContainElement(&iri.NetworkInterface{
 				Name: "nic-1",
 			})),
-			HaveField("Machine.Status.ObservedGeneration", BeZero()),
+			HaveField("Machine.Status.ObservedGeneration", Equal(int64(1))),
```
lukas016, line 403: FYI (pls don't solve it in this PR): i checked logic and it looks little weird for me, we return api.Machine.Metadata.Generation as observedGeneration. But it isn't true because worker can be overloaded  or stuck somewhere and we don't have guaranty, if desired state state was processed. So technically we are reporting invalid information to customer.

pkg/host/store.go
so-sahu: Your suggestion to avoid using the generation field and instead extend metadata with resourceVersion aligns with Kubernetes API concepts and is a valid approach in this scenario. 

Concurrency control and consistency in Kubernetes are achieved through several mechanisms, including the use of resourceVersion. More details on how Kubernetes handles concurrency control and consistency can be found in its API conventions document: [Concurrency Control and Consistency in Kubernetes API](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency).

pkg/host/store.go
```diff
@@ -129,6 +130,11 @@ func (s *Store[E]) Update(ctx context.Context, obj E) (E, error) {
 		return obj, nil
 	}
 
+	if oldObj.GetGeneration() != obj.GetGeneration() {
```
so-sahu, line 133: I believe it would be better to perform this check before deepEqual. If the check fails, it indicates that the file in the store has already been modified, and no further operations with this object are necessary for the current reconciliation.

so-sahu (changes requested):

pkg/host/store.go
kasabe28: Done

provider/server/machine_create_test.go
kasabe28: Reverted as no longer needed

pkg/host/store.go
kasabe28: Done

pkg/host/store.go
kasabe28: Done

pkg/host/store.go
kasabe28: used IncrementResourceVersion()

lukas016 (dismissed):

pkg/api/api.go
```diff
@@ -10,9 +10,10 @@ type Metadata struct {
 	Annotations map[string]string `json:"annotations"`
 	Labels      map[string]string `json:"labels"`
 
-	CreatedAt  time.Time  `json:"createdAt"`
-	DeletedAt  *time.Time `json:"deletedAt,omitempty"`
-	Generation int64      `json:"generation"`
+	CreatedAt       time.Time  `json:"createdAt"`
+	DeletedAt       *time.Time `json:"deletedAt,omitempty"`
+	Generation      int64      `json:"generation"`
+	ResourceVersion uint64     `json:"resourceVersion,omitempty"`
```
so-sahu, line 16: The ResourceVersion is always incremented from the creation process, so it can never be 0. If it ever appears as 0, it indicates an abnormal scenario and should be explicitly visible. Therefore, the omitempty parameter is not necessary in this case.

kasabe28:
> Root cause wasn't solved

It should solve it now.

pkg/host/store.go
```diff
@@ -128,10 +129,16 @@ func (s *Store[E]) Update(_ context.Context, obj E) (E, error) {
 		return obj, nil
 	}
 
+	if oldObj.GetResourceVersion() != obj.GetResourceVersion() {
+		return utils.Zero[E](), fmt.Errorf("failed to update object: resourceVersion is not latest")
```
lukas016, line 133: could you please create exporter error which can be used in other parts of code for detect this problem.

```
var ErrResourceVersionNotLatest = errors.New("resourceVersion is not latest")

fmt.Errorf("failed to update object: %w, ErrResourceVersionNotLatest)
```

lukas016 (dismissed):
LGTM

pkg/host/store.go
```diff
@@ -128,10 +129,16 @@ func (s *Store[E]) Update(_ context.Context, obj E) (E, error) {
...
+		return utils.Zero[E](), fmt.Errorf("failed to update object: %w", store.ErrResourceVersionNotLatest)
+	}
+
 	if reflect.DeepEqual(oldObj, obj) {
 		return obj, nil
 	}
 
+	obj.IncrementResourceVersion()
+
 	//Todo: update version
```
so-sahu, line 142: I think this comment can now be removed.

lukas016 (approved):
LGTM

afritzler (approved):

PR #196, opened 2024-02-20
Implemented metrics endpoint

### LINKED ISSUE #57: Implement support for metrics server
# Summary

We need for monitoring implement support of Prometheus metrics. The same functionality was before implemented directly in controller-runtime manager.

You can read more about implementation here: https://prometheus.io/docs/guides/go-application/
 We need implement only base support for metrics in this issue. Implementation of custom metrics will do in another issue.

Metrics server has to be optional and it can be enabled with arguments

## Motivation

Rise quality of service

lukas016:
# Proposed Changes

- implement standard prometheus metric server

Fixes #57

### DIFF
--- a/provider/cmd/app/app.go
+++ b/provider/cmd/app/app.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -92,10 +93,11 @@ type Options struct {
 }
 
 type HTTPServerOptions struct {
-	Addr         string
-	ReadTimeout  time.Duration
-	WriteTimeout time.Duration
-	IdleTimeout  time.Duration
+	Addr            string
+	ReadTimeout     time.Duration
+	WriteTimeout    time.Duration
+	IdleTimeout     time.Duration
+	GracefulTimeout time.Duration
 }
 
 type ServersOptions struct {
@@ -129,7 +131,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Servers.Metrics.Addr, "servers-metrics-address", "", "Address to listen on exposing of metrics. If address isn't set, server is disabled.")
 	fs.DurationVar(&o.Servers.Metrics.ReadTimeout, "servers-metrics-readtimeout", 200*time.Millisecond, "Read timeout for metrics server.")
 	fs.DurationVar(&o.Servers.Metrics.WriteTimeout, "servers-metrics-writetimeout", 200*time.Millisecond, "Write timeout for metrics server.")
-	fs.DurationVar(&o.Servers.Metrics.IdleTimeout, "server-metrics-idletimeout", 1*time.Second, "Idle timeout for connections to metrics server.")
+	fs.DurationVar(&o.Servers.Metrics.IdleTimeout, "servers-metrics-idletimeout", 1*time.Second, "Idle timeout for connections to metrics server.")
+	fs.DurationVar(&o.Servers.Metrics.GracefulTimeout, "servers-metrics-gracefultimeout", 2*time.Second, "Graceful timeout for shutdown metrics server. Ideally set it little longer as idletimeout.")
 
 	fs.BoolVar(&o.EnableHugepages, "enable-hugepages", false, "Enable using Hugepages.")
 	fs.Var(&o.GuestAgent, "guest-agent-type", fmt.Sprintf("Guest agent implementation to use. Available: %v", guestAgentOptionAvailable()))
@@ -496,22 +499,30 @@ func runMetricsServer(ctx context.Context, setupLog logr.Logger, opts HTTPServer
 		IdleTimeout:  opts.IdleTimeout,
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		<-ctx.Done()
-		setupLog.Info("Shutting down streaming server")
-		shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+		setupLog.Info("Shutting down metrics server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), opts.GracefulTimeout)
 		defer cancel()
 		locErr := srv.Shutdown(shutdownCtx)
 		if locErr != nil {
 			setupLog.Error(locErr, "metrics server wasn't shutdown properly")
+		} else {
+			setupLog.Info("Metrics server is shutdown")
 		}
-		setupLog.Info("Metrics server is shutdown")
 	}()
 
 	err := srv.ListenAndServe()
 	if err != nil && !errors.Is(err, http.ErrServerClosed) {
 		return fmt.Errorf("error listening / serving metrics server: %w", err)
 	}
+
+	setupLog.Info("Metrics server stopped serve new connections")
+
+	wg.Wait()
 
 	return nil
 }
--- a/provider/cmd/app/app.go
+++ b/provider/cmd/app/app.go
@@ -101,6 +101,7 @@ type HTTPServerOptions struct {
 type ServersOptions struct {
 	Metrics HTTPServerOptions
 }
+
 type LibvirtOptions struct {
 	Socket  string
 	Address string
@@ -125,7 +126,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.BaseURL, "base-url", "", "The base url to construct urls for streaming from. If empty it will be "+
 		"constructed from the streaming-address")
 
-	fs.StringVar(&o.Servers.Metrics.Addr, "servers-metrics-address", ":8080", "Address to listen on exposing of metrics.")
+	fs.StringVar(&o.Servers.Metrics.Addr, "servers-metrics-address", "", "Address to listen on exposing of metrics. If address isn't set, server is disabled.")
 	fs.DurationVar(&o.Servers.Metrics.ReadTimeout, "servers-metrics-readtimeout", 200*time.Millisecond, "Read timeout for metrics server.")
 	fs.DurationVar(&o.Servers.Metrics.WriteTimeout, "servers-metrics-writetimeout", 200*time.Millisecond, "Write timeout for metrics server.")
 	fs.DurationVar(&o.Servers.Metrics.IdleTimeout, "server-metrics-idletimeout", 1*time.Second, "Idle timeout for connections to metrics server.")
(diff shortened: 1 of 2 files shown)


provider/cmd/app/app.go
```diff
@@ -455,3 +464,37 @@ func runStreamingServer(ctx context.Context, setupLog, log logr.Logger, srv *ser
...
+func runMetricsServer(ctx context.Context, setupLog logr.Logger, addr string) error {
+	setupLog.Info("Starting metrics server on " + addr)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+
+	srv := http.Server{
+		Addr:         addr,
+		Handler:      mux,
+		ReadTimeout:  5 * time.Second,
```
so-sahu, line 477: Is it a good idea to make these timeouts configurable?

provider/cmd/app/app.go
```diff
@@ -455,3 +464,37 @@ func runStreamingServer(ctx context.Context, setupLog, log logr.Logger, srv *ser
...
+		IdleTimeout:  2 * time.Second,
+	}
+
+	go func() {
+		<-ctx.Done()
+		setupLog.Info("Shutting down streaming server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+		locErr := srv.Shutdown(shutdownCtx)
+		if locErr != nil {
```
kasabe28, line 488: optional: I would write line 487 & 488 as:
`if locErr := srv.Shutdown(shutdownCtx); locErr != nil {`

same for line 494 & 495

lukas016: i personally don't like this style. Juniors have problem understand life scope of variable properly (personally tested) and you can very easy do shadowing what is just more confusing.
+ i saw very long if statements with this style which are worse readable.
So i don't use this short format.

provider/cmd/app/app.go
lukas016: it is good idea, but i did it for streaming server and maybe grpc too. So what do you think about follow up issue?

provider/cmd/app/app.go
kasabe28: I don't think so its worse readable. And plus point is variables declared by the statement are only in scope until the end of the if. Anyway it was optional.

provider/cmd/app/app.go
so-sahu: It appears that there are no timeout parameters specified in the gRPC server. Additionally, in the current implementations, timeout settings are not defined for the streaming server, which implies that there will never be a timeout. The question is, are we considering introducing timeouts for the streaming server?

I would suggest making timeouts for the metrics server here and creating issues for other components if we plan to do so.

lukas016: GRPC has timeout: https://pkg.go.dev/google.golang.org/grpc#ConnectionTimeout

timeouts are for protection of our app. because you can do attacks with long living connections. 5 *time.Second is extremely long time.
but i will do it

lukas016: fixed

provider/cmd/app/app.go
```diff
@@ -88,6 +91,16 @@ type Options struct {
...
+type HTTPServerOptions struct {
+	Addr         string
+	ReadTimeout  time.Duration
+	WriteTimeout time.Duration
+	IdleTimeout  time.Duration
+}
+
+type ServersOptions struct {
+	Metrics HTTPServerOptions
+}
```
lukas016, line 103: add new line

provider/cmd/app/app.go
```diff
@@ -114,6 +127,11 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.VirshExecutable, "virsh-executable", "virsh", "Path / name of the virsh executable.")
 
+	fs.StringVar(&o.Servers.Metrics.Addr, "servers-metrics-address", ":8080", "Address to listen on exposing of metrics.")
```
lukas016, line 130: if address isn't set, server will be disabled

provider/cmd/app/app.go
```diff
@@ -349,6 +368,11 @@ func Run(ctx context.Context, opts Options) error {
 	}
 
 	g, ctx := errgroup.WithContext(ctx)
+
+	g.Go(func() error {
```
so-sahu, line 372: Should this function still be run in a goroutine if the metrics server is disabled?

provider/server/server_suite_test.go
```diff
@@ -99,6 +100,11 @@ var _ = BeforeSuite(func() {
 		PathSupportedMachineClasses: machineClassesFile.Name(),
 		RootDir:                     filepath.Join(tempDir, "libvirt-provider"),
 		StreamingAddress:            streamingAddress,
+		Servers: app.ServersOptions{
+			Metrics: app.HTTPServerOptions{
+				Addr: "127.0.0.1:9090",
```
so-sahu, line 105: Optional: Timeouts can be specified here as well.

provider/server/server_suite_test.go
```diff
@@ -99,6 +100,11 @@ var _ = BeforeSuite(func() {
 		PathSupportedMachineClasses: machineClassesFile.Name(),
 		RootDir:                     filepath.Join(tempDir, "libvirt-provider"),
 		StreamingAddress:            streamingAddress,
+		Servers: app.ServersOptions{
```
so-sahu, line 103: If we're not conducting any specific integration tests for metrics, is there a need to include this here?

provider/cmd/app/app.go
lukas016: go routines will end and program still is running. I don't see any big problem with that and i have code in one function.

provider/server/server_suite_test.go
lukas016: i will disable metrics server for integration test.

provider/server/server_suite_test.go
lukas016: options has to be included but it can be disabled.

provider/server/server_suite_test.go
lukas016: fixed

provider/server/server_suite_test.go
lukas016: fixed

lukas016: optional isn't pointer so we don't have to defined it here. But i prefer explicit configuration, developer can immediately see how we run app without study code or docs.

provider/cmd/app/app.go
```diff
@@ -453,3 +477,42 @@ func runStreamingServer(ctx context.Context, setupLog, log logr.Logger, srv *ser
...
+	go func() {
+		<-ctx.Done()
+		setupLog.Info("Shutting down streaming server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+		locErr := srv.Shutdown(shutdownCtx)
+		if locErr != nil {
+			setupLog.Error(locErr, "metrics server wasn't shutdown properly")
+		}
+		setupLog.Info("Metrics server is shutdown")
```
so-sahu, line 509: Here, even if the Shutdown function returns an error, the statement following the error handling will still be executed. This could be misleading as it implies that the server shutdown was successful, even if an error occurred.

so-sahu (changes requested):

provider/cmd/app/app.go
lukas016: i will move it to line 516, it will be better

lukas016: fixed. I did shutdown and logging more flexible.

provider/cmd/app/app.go
```diff
@@ -497,22 +500,30 @@ func runMetricsServer(ctx context.Context, setupLog logr.Logger, opts HTTPServer
 		IdleTimeout:  opts.IdleTimeout,
 	}
 
+	var wg sync.WaitGroup
+	wg.Add(1)
```
so-sahu, line 504: Since only one goroutine is being spawned, there's no need for the wg WaitGroup. Adding wg.Add(1) and wg.Wait() for a single goroutine introduces unnecessary complexity.

provider/cmd/app/app.go
```diff
@@ -497,22 +500,30 @@ func runMetricsServer(ctx context.Context, setupLog logr.Logger, opts HTTPServer
...
 		}
-		setupLog.Info("Metrics server is shutdown")
 	}()
 
 	err := srv.ListenAndServe()
 	if err != nil && !errors.Is(err, http.ErrServerClosed) {
 		return fmt.Errorf("error listening / serving metrics server: %w", err)
 	}
 
+	setupLog.Info("Metrics server stopped serve new connections")
```
so-sahu, line 524: Optional: This log statement appears unnecessary as it provides more of an informational message rather than being critical for the operation. Additionally, since we are already logging at line 508, it may not be very helpful in this context.

so-sahu (changes requested):
Can be simplified.

provider/cmd/app/app.go
lukas016: how do you mean that? I don't want to close parent goroutine before end goroutine for shutdown.

provider/cmd/app/app.go
lukas016: I would like keep it here. it can be critical, because the shutdown process will stop accepting new connections, but the server can still live and serve existing connections, depending on the configuration graceful period. So you can have this weird state, when you cannot get metrics anymore but server is still running.

provider/cmd/app/app.go
so-sahu: ok. I agree with you.  If the WaitGroup is not used to wait for this goroutine to finish, the runMetricsServer function could return before the metrics server is properly shut down, potentially leaving the server running in the background.

so-sahu (dismissed):
LGTM

so-sahu (dismissed):
LGTM

provider/cmd/app/app.go
```diff
@@ -88,6 +92,18 @@ type Options struct {
 	ResyncIntervalGarbageCollector time.Duration
 }
 
+type HTTPServerOptions struct {
+	Addr            string
+	ReadTimeout     time.Duration
+	WriteTimeout    time.Duration
+	IdleTimeout     time.Duration
```
lukasfrank, line 99: Can we go simple here? I propose to extend the options if we see any issues and drop it for now.

lukas016: why? it is configurable, you can redefine it, if you need it or use default value. But you have this option and you don't have to wait for implementation. It can be helpful because you don't want to have multiple long living connection because you have limited count of filedescriptors which can be open at the time. So indirectly connection can affects main logic.

lukas016: fixed

lukasfrank (approved):

PR #200, opened 2024-02-22
Add test cases for serverExec

### LINKED ISSUE #90: Provide tests for `ServeExec`
# Summary

A test for `ServeExec` should be provided to cover at least: 
- that a http request with a not expected method fails
- that a request with a not existing token fails
- that a token can't be used twice
- that a request with a valid token and a not existing machine fails

This should be implemented using `net/http/httptest`. 

_Originally posted by @lukasfrank in https://github.com/ironcore-dev/libvirt-provider/pull/65#discussion_r1432477420_

ushabelgur:
# Proposed Changes

- add test  for validating http request with a not expected method fails
- add test for validating that a token can't be used twice
- add test to validate a request with a valid token and a not existing machine fails

Fixes #90


provider/server/exec_test.go
```diff
@@ -68,5 +67,69 @@ var _ = Describe("Exec", func() {
...
 	})
 })
+
+func runExec(ctx context.Context, execUrl *url.URL) error {
+	randomSize := 1024 * 1024
+	randomData := make([]byte, randomSize)
+	var stdout bytes.Buffer
+
+	tty := term.TTY{
+		In:     bytes.NewReader(randomData),
```
lukas016, line 109: what are you testing here?

provider/server/exec_test.go
```diff
@@ -68,5 +67,69 @@ var _ = Describe("Exec", func() {
...
+
+		By("issuing exec for non existing machine")
+		err = runExec(ctx, parsedResUrl)
+		machineNotSynchErr := fmt.Sprintf("machine %s has not yet been synced", createResp.Machine.Metadata.Id)
+		Expect(err).To(SatisfyAny(MatchError(ContainSubstring("404 page not found")), MatchError(ContainSubstring(machineNotSynchErr))))
+
 	})
 })
+
+func runExec(ctx context.Context, execUrl *url.URL) error {
```
lukas016, line 103: this function doesn't try send any command to console or read output. So you don't test whole logic of Streaming server. You test just http server logic but i recommend test whole streaming logic with send command and read output. Maybe login of user will enough.

lukas016 (changes requested):
machine isn't deleted, when test fail

provider/server/exec_test.go
```diff
@@ -29,15 +37,6 @@ var _ = Describe("Exec", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(createResp).NotTo(BeNil())
 
-		DeferCleanup(func(ctx SpecContext) {
```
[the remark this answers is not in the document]
ushabelgur: I moved delete to end of the test case to cover to this case: to validate a request with a valid token and a not existing machine fails

provider/server/exec_test.go
ushabelgur: It was not mentioned in the scope of issue https://github.com/ironcore-dev/libvirt-provider/issues/90 If really i can cover. It talks about only negative cases.

provider/server/exec_test.go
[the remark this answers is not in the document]
lukas016: if test fail, machine will still exist what can affect another tests. So you need properly cleanup env after end of test and it doesn't matter on result of test.

provider/server/exec_test.go
lukas016: we can discuss it on daily. But implement whole logic can help us catch error during development later.

provider/server/exec_test.go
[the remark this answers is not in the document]
ushabelgur: Added back deletion part in DefereCleanup

hardikdr (commented):
Thanks for the PR @ushabelgur .

Can you please inform here when the requested changes has already been taken care of.

ushabelgur:
@hardikdr Machine deletion part at the end of test case has been addressed.  But the test case to validate whole streaming logic with send command and read output, I am still working on it.

provider/server/exec_test.go
ushabelgur: We are using go routine with infinite loop to read and write to stream, even when i am sending Escape Sequence to terminate the stream it is not coming out of the go routine. Tried with bytes.Buffer Read and Write methods, somehow escape sequence is not working with this kind of buffer. I have created separate issue to work on it further #207

provider/server/exec_test.go
```diff
@@ -30,9 +38,12 @@ var _ = Describe("Exec", func() {
 		Expect(createResp).NotTo(BeNil())
 
 		DeferCleanup(func(ctx SpecContext) {
-			_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
-			Expect(err).ShouldNot(HaveOccurred())
 			Eventually(func() bool {
+				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
```
lukas016, line 42: pls call DeleteMachine just one time before Eventually

ushabelgur: We can not take it out of Eventually, we were facing this machine object getting overridden problem sometimes as mentioned in #164 . In all other integration test also its been handled similar way. May be after this fix we can tryout.

lukas016 (dismissed):
LGTM, tests will improve in follow up issue: https://github.com/ironcore-dev/libvirt-provider/issues/207

lukas016 (approved):
LGTM

hardikdr (approved):
Thanks for the PR @ushabelgur, appreciate it.

PR #203, opened 2024-02-23
Adding Resource Fields for `api.Machine`

### LINKED ISSUE #181: Extending `api.Machine` with needed properties
# Summary
This issue captures the efforts to extend the API machine object and the corresponding reconciler changes. The proposals and changes can be discussed in this issue.

### LINKED ISSUE #199: Simplify Setting Default Value for Flag `volume-size-resync-duration`
**Description**
In the current implementations, the default value for flag `volume-size-resync-duration` is set to zero value [here](https://github.com/ironcore-dev/libvirt-provider/blob/80ef3fbc7d797805eafbba7b43e755e8a1285184/provider/cmd/app/app.go#L107) and later on in the machine controller it's been set to 1 minute [here](https://github.com/ironcore-dev/libvirt-provider/blob/80ef3fbc7d797805eafbba7b43e755e8a1285184/pkg/controllers/machine_controller.go#L78). The default value can directly be set in the flag initialization which can simplify the flow and avoid unnecessary logic.

**Proposed Solution**
Update the flag initialization for `volume-size-resync-duration` in the libvirt-provider to directly set the default value to 1 minute. This will simplify the code and make it more straightforward.

**Additional context**
This change simplifies the code and makes it more consistent with the expected behavior.

so-sahu:
Fixes #181

### DIFF
--- a/pkg/api/machine.go
+++ b/pkg/api/machine.go
@@ -18,8 +18,8 @@ type Machine struct {
 type MachineSpec struct {
 	Power PowerState `json:"power"`
 
-	CpuMillis   int64 `json:"cpuMillis"`
-	MemoryBytes int64 `json:"memoryBytes"`
+	Resources       ResourceList    `json:"resources"`
+	NUMAPreferences NUMAPreferences `json:"numaPreferences,omitempty"`
 
 	Image    *string `json:"image"`
 	Ignition []byte  `json:"ignition"`
@@ -31,6 +31,7 @@ type MachineSpec struct {
 }
 
 type MachineStatus struct {
+	NUMAPlacement          NUMAPlacement            `json:"numaPlacement,omitempty"`
 	VolumeStatus           []VolumeStatus           `json:"volumeStatus"`
 	NetworkInterfaceStatus []NetworkInterfaceStatus `json:"networkInterfaceStatus"`
 	State                  MachineState             `json:"state"`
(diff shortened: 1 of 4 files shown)


pkg/controllers/machine_controller.go
```diff
@@ -765,7 +765,7 @@ func (r *MachineReconciler) setDomainResources(machine *api.Machine, domain *lib
 		}
 	}
 
-	cpu := uint(machine.Spec.CpuMillis / 1000)
+	cpu := uint(machine.Spec.Resources.CPU().Value() / 1000)
```
lukas016, line 768: this logic is wrong. 8.6 will rounded down to 8 what doesn't make sense for me. I prefer rounded up.

@afritzler do we need support millis unit? Cannot we support just whole cores? 1, 2, 3, 4, 5, 6, 7, 8, ...

lukas016 (changes requested):

pkg/api/machine.go
```diff
@@ -31,6 +31,7 @@ type MachineSpec struct {
 }
 
 type MachineStatus struct {
+	NUMAPlacement          NUMAPlacement            `json:"numaPlacement,omitempty"`
```
lukas016, line 34: here maybe you will need to add ResourceList too

pkg/api/resource.go
```diff
@@ -0,0 +1,57 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package api
+
+import "k8s.io/apimachinery/pkg/api/resource"
+
+type ResourceName string
+
+const (
+	ResourceCPU       ResourceName = "cpu"
```
lukas016, line 11: pls don't create own ResourceName or type. Ideally reuse core package: core "github.com/ironcore-dev/ironcore/api/core/v1alpha1"

Issue #118, opened 2024-01-17
Add Size Tolerance for Encrypted Volumes During Resize


hardikdr:
# Summary

A slight variation in the size of Encrypted volumes is observed when attached to the VMs. This could be as small as 0.2% less than the attached volume's size. 
This can confuse the volume resize controller to trigger the resize operation, hence either a small percent or an absolute value equivalent of the difference should be ignored while reconciling. 

## Motivation

Better robustness.

sujeet01:
@hardikdr I haven't noticed any size differences in our `Microceph` test workflow. It's possible that `Microceph` doesn't support RBD image encryption. 

Could you please confirm if you're still observing size difference in encrypted volume attachment on your side?"

PR #213, opened 2024-03-01
Enhance volume attachment integration tests using `ceph-provider` volume client

sujeet01:
Enhance volume attachment integration tests using `ceph-provider` volume client for
1. non-encrypted volume attachment
2. encrypted volume attachment


provider/server/machine_volume_attach_test.go
```diff
@@ -90,25 +103,301 @@ var _ = Describe("AttachVolume", func() {
...
+					Class: machineClassx2medium,
+				},
+			},
+		})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(createResp).NotTo(BeNil())
+
+		DeferCleanup(func(ctx SpecContext) {
+			Eventually(func() bool {
+				_, err := machineClient.DeleteMachine(ctx, &machineiriv1alpha1.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
```
lukas016, line 299: unclear using of ctx

lukas016: optional: rename err to locErr or defErr 

i am not big fan of shadowing. Juniors have problem with it and it can very easy change behavior of program when you delete one character ':', so it affects review process because when i see shadowing i have to me careful with error flow

provider/server/machine_volume_attach_test.go
```diff
@@ -36,7 +37,7 @@ var _ = Describe("AttachVolume", func() {
 
 		DeferCleanup(func(ctx SpecContext) {
 			Eventually(func() bool {
-				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+				_, err := machineClient.DeleteMachine(ctx, &machineiriv1alpha1.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
 				Expect(err).To(SatisfyAny(
 					BeNil(),
 					MatchError(ContainSubstring("NotFound")),
```
lukas016, line 43: i recommend extend this test with check of filesystem. We will need to check, if folder for machine is deleted when machine was deleted. We have this logic in different test.

lukas016 (changes requested):

provider/server/machine_volume_attach_test.go
```diff
@@ -7,26 +7,27 @@ import (
...
+		createResp, err := machineClient.CreateMachine(ctx, &machineiriv1alpha1.CreateMachineRequest{
+			Machine: &machineiriv1alpha1.Machine{
+				Metadata: &irimetav1alpha1.ObjectMetadata{
 					Labels: map[string]string{
 						"foo": "bar",
 					},
 				},
-				Spec: &iri.MachineSpec{
-					Power: iri.Power_POWER_ON,
+				Spec: &machineiriv1alpha1.MachineSpec{
```
lukas016, line 29: i recommend use squashfs and properly booted os. it is more close to our production environment. Here are you doing test with "unsupported" setup.

sujeet01:
Marking this PR as draft as it seems that Microceph lacks support for encrypting rbd images. Further investigation is required.

sujeet01:
Closing this PR because the size difference is no longer observed https://github.com/ironcore-dev/libvirt-provider/issues/118 making the test enhancement proposed in this PR unnecessary.

PR #222, opened 2024-03-06
check boot of vm with ssh connection

### LINKED ISSUE #217: Use `Eventually` and try to wait until the SSH port of the machine is open
Sleep is good enough for now but is not the best solution here. Alternatively we could think of using `Eventually` and try to wait until the SSH port of the machine is open. We should be able to use a local network device as a `NetworkInterface` here.

_Originally posted by @afritzler in https://github.com/ironcore-dev/libvirt-provider/pull/189#discussion_r1510764149_

lukas016:
# Proposed Changes

Fixes #217


lukas016:
i did over 300 runs and every run was successful. i couldn't simulate problem with getting private IP. It looks as env problem.

lukas016:
PR can be merge after PR https://github.com/ironcore-dev/libvirt-provider/pull/245. Some small changes for improve stability from this PR will be implemented there.

PR #223, opened 2024-03-06
stop rewrite DeletedAt for every call

lukas016:
# Proposed Changes

DeletdeAt in api.Machine was rewrite for every iri call DeleteMachine. It created conflicts with other components which trying update api.Machine object and information when deleteMachine call was called first time was lost.

### DIFF
--- a/pkg/host/store.go
+++ b/pkg/host/store.go
@@ -165,14 +165,16 @@ func (s *Store[E]) Delete(_ context.Context, id string) error {
 		return s.delete(id)
 	}
 
-	if obj.GetDeletedAt() == nil {
-		now := time.Now()
-		obj.SetDeletedAt(&now)
-		obj.IncrementResourceVersion()
+	if obj.GetDeletedAt() != nil {
+		return nil
+	}
 
-		if _, err := s.set(obj); err != nil {
-			return fmt.Errorf("failed to set object metadata: %w", err)
-		}
+	now := time.Now()
+	obj.SetDeletedAt(&now)
+	obj.IncrementResourceVersion()
+
+	if _, err := s.set(obj); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/host/store.go
```diff
@@ -165,12 +165,14 @@ func (s *Store[E]) Delete(_ context.Context, id string) error {
...
+		obj.IncrementResourceVersion()
 
-	if _, err := s.set(obj); err != nil {
-		return fmt.Errorf("failed to set object metadata: %w", err)
+		if _, err := s.set(obj); err != nil {
+			return fmt.Errorf("failed to set object metadata: %w", err)
+		}
 	}
 
 	s.enqueue(store.WatchEvent[E]{
```
so-sahu, line 178: If a second delete request is sent, nothing happens in this code. Do we still need to trigger the delete event in this scenario?

lukas016: i don't know, so i kept previous logic. Minimally it will still trigger reconcile loop.

so-sahu: It doesn't seem necessary to trigger the reconcile loop if there are no changes.

lukas016: it is the design problem, you don't retrigger reconcile loop only but you notified every goroutines which registered watch function. So i kept this logic. @lukasfrank you will know better. I agree with @so-sahu and it isn't needed probably but i wanted risk side effects with designed change.

lukas016: you have right, i checked Update function and when object doesn't contain changes, it doesn't send event. I will change it.

lukas016: fixed

so-sahu (approved):
LGTM

hardikdr (approved):

PR #228, opened 2024-03-08
Removing `virsh` dependency

### LINKED ISSUE #71: Remove Dependency from the `Virsh`
# Summary

This is an issue to discuss and explore the feasibility of removing the dependency on the `virsh` CLI in the implementation of Machine Exec functionality.

One such alternative is the `DomainSerialConsole()` API provided by DigitalOcean's `go-libvirt` library. However, this API is unidirectional, which might not fully cater to our requirements. The potential helper there is using `DomainSendKeys` for sending KeyCodes, but this approach seems to be cumbersome, keyboard-type/os-specific, especially for the start.

The purpose here is to initiate a discussion and research into the available options for the future.

## Motivation

Better maintainability, and less dependency on the host.

so-sahu:
# Proposed Changes

- Using Pseudo-Terminal (pty) for bi-directional communication
- Created development setup document and mentioned about the needed access.

Fixes https://github.com/ironcore-dev/libvirt-provider/issues/71

### DIFF
--- a/docs/development/setup.md
+++ b/docs/development/setup.md
@@ -82,24 +82,24 @@
     Sample `iri-machine.yaml`:
 
     ```yaml
-        metadata:
-          id: 91076287116041d00fd421f43c3760389041dac4a8bd9201afba9a5baeb21c7
-          labels:
-            downward-api.machinepoollet.api.onmetal.de/root-machine-name: machine-hd4
-            downward-api.machinepoollet.api.onmetal.de/root-machine-namespace: default
-            downward-api.machinepoollet.api.onmetal.de/root-machine-uid: cab82eac-09d8-4428-9e6c-c98b40027b74
-            machinepoollet.api.onmetal.de/machine-name: machine-hd4
-            machinepoollet.api.onmetal.de/machine-namespace: default
-            machinepoollet.api.onmetal.de/machine-uid: cab82eac-09d8-4428-9e6c-c98b40027b74
-        spec:
-          class: x3-small
-          image:
-            image: ghcr.io/ironcore-dev/ironcore-image/gardenlinux:rootfs-dev-20231206-v1
-          volumes:
-          - empty_disk:
-              size_bytes: 5368709120
-            name: ephe-disk
-            device: oda
+    metadata:
+      id: 91076287116041d00fd421f43c3760389041dac4a8bd9201afba9a5baeb21c7
+      labels:
+        downward-api.machinepoollet.api.onmetal.de/root-machine-name: machine-hd4
+        downward-api.machinepoollet.api.onmetal.de/root-machine-namespace: default
+        downward-api.machinepoollet.api.onmetal.de/root-machine-uid: cab82eac-09d8-4428-9e6c-c98b40027b74
+        machinepoollet.api.onmetal.de/machine-name: machine-hd4
+        machinepoollet.api.onmetal.de/machine-namespace: default
+        machinepoollet.api.onmetal.de/machine-uid: cab82eac-09d8-4428-9e6c-c98b40027b74
+    spec:
+      class: x3-small
+      image:
+        image: ghcr.io/ironcore-dev/ironcore-image/gardenlinux:rootfs-dev-20231206-v1
+      volumes:
+      - empty_disk:
+          size_bytes: 5368709120
+        name: ephe-disk
+        device: oda
     ```
 
 1. **Listing machines**
--- a/docs/development/setup.md
+++ b/docs/development/setup.md
@@ -7,7 +7,7 @@
 - [Deploy `libvirt-provider`](#deploy-libvirt-provider)
 
 > ℹ️ **NOTE**:</br>
-> To be able to take exec console of the machine, you can follow either of the below approaches:</br>
+> To be able to take exec console of the machine, you can follow any one of the below approaches:</br>
 > - Run the `libvirt-provider` as the `libvirt-qemu` user.</br>
 > - Create an entry with `devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=0660 0 0` in `/etc/fstab`.</br>
 > - Manually ensure that you have `0660` access permissions on the character files created in `/dev/pts`.</br>
@@ -17,7 +17,6 @@
 - Linux (code contains OS specific code)
 - go >= 1.20
 - `git`, `make` and `kubectl`
-- [Kustomize](https://kustomize.io/)
 - Access to a Kubernetes cluster ([Minikube](https://minikube.sigs.k8s.io/docs/), [kind](https://kind.sigs.k8s.io/) or a
   real cluster)
 - [libvirt](http://libvirt.org)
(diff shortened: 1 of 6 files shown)


docs/development/setup.md
```diff
@@ -0,0 +1,156 @@
...
+- [Prerequisites](#prerequisites)
+- [Preperation](#preperation)
+- [Run libvirt-provider for local development](#run-libvirt-provider-for-local-development)
+- [Interact with the `libvirt-provider`](#interact-with-the-libvirt-provider)
+- [Deploy `libvirt-provider`](#deploy-libvirt-provider)
+
+> ℹ️ **NOTE**:</br>
+> To be able to take exec console of the machine, you can follow either of the below approaches:</br>
+> - Run the `libvirt-provider` as the `libvirt-qemu` user.</br>
+> - Create an entry with `devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=0660 0 0` in `/etc/fstab`.</br>
```
lukas016, line 12: is it needed, if libvirt-provider runs as libvirt-qemu user?

docs/development/setup.md
```diff
@@ -0,0 +1,156 @@
...
+
+```bash
+go run provider/cmd/main.go \
+   --libvirt-provider-dir=<path-to-initialize-libvirt-provider> \
+   --supported-machine-classes=<path-to-machine-class-json>/machine-classes.json \
+   --network-interface-plugin-name=isolated \
+   --address=<local-path>/iri-machinebroker.sock
+```
+
+Sample `machine-classes.json`:
```
lukas016, line 73: sample is available in  [machineclasses.json](https://github.com/ironcore-dev/libvirt-provider/blob/main/config/development/machineclasses.json)

provider/server/exec.go
```diff
@@ -149,7 +153,7 @@ func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser
 			n, err := inputReader.Read(buf)
 			if err != nil {
 				if _, ok := err.(term.EscapeError); ok {
-					_, _ = f.Write(buf) // This is to close the writer.
+					f.Close() // This is to close the writer.
```
lukas016, line 156: why is it here? i expect it on end of function  behind wg.Wait

docs/development/setup.md
```diff
@@ -0,0 +1,156 @@
...
+- [QEMU](https://www.qemu.org/download/)
+- `irictl-machine` should be running locally or as [container](https://github.com/ironcore-dev/ironcore/pkgs/container/ironcore-irictl-machine)
+
+## Preperation
+
+### Setup `irictl-machine`
+
+**1. Clone ironcore repository**
+
+```bash
```
lukas016, line 33: i think this can be do better

provider/server/exec.go
```diff
@@ -124,16 +123,21 @@ func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser
...
+		return fmt.Errorf("failed to lookup domain: %w", err)
 	}
 
-	cmd := exec.CommandContext(ctx, e.VirshExecutable, "-c", strings.TrimSpace(uri), "console", machineID)
+	domainXML := &libvirtxml.Domain{}
+	if err := domainXML.Unmarshal(domainXMLData); err != nil {
+		return fmt.Errorf("failed to unmarshal domain: %w", err)
+	}
+
+	ttyPath := domainXML.Devices.Consoles[0].TTY
```
lukas016, line 136: i recommend check length of Consoles array, this can create panic.

provider/server/exec.go
lukas016: or ideally use defer

lukas016 (changes requested):
code can be written more secure

docs/development/setup.md
so-sahu: As stated in line 10, you can choose either of the approaches listed. I have included all the options here because we have not yet reached a consensus on which one to follow.

docs/development/setup.md
so-sahu: Okay, I'll add the link here. Thank you for bringing it to my attention.

provider/server/exec.go
so-sahu: The current position of f.Close() is appropriate, as it ensures that the file descriptor is closed at the right moment, when the escape sequence is received. Moving f.Close() to a defer statement within the goroutine would not work as intended. Deferring f.Close() would mean that it is called when the goroutine exits, which is not desirable in this case. The f.Close() call needs to be executed immediately when the escape sequence is detected to ensure that the writer (the goroutine reading from in) exits the infinite loop and the file descriptor is closed promptly.

docs/development/setup.md
so-sahu: Could you please give more context here.

provider/server/exec.go
so-sahu: Since we set this exclusively in the domainXML during machine creation, it won't cause a panic.

docs/development/setup.md
lukas016: you have right, my fault

docs/development/setup.md
lukas016: i want to show you it offline because it is more changes.

provider/server/exec.go
lukas016: i prefer secure programming here. logic code can be change in future and new developer doesn't have to know about this code. It will throw panic which can restart app or it will cache be server. I don't know current implementation and i don't want to risk restart of whole app.

provider/server/exec.go
```diff
@@ -124,16 +123,21 @@ func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser
 		return fmt.Errorf("machine %s has not yet been synced", machineID)
 	}
 
-	uri, err := e.Libvirt.ConnectGetUri()
+	domainXMLData, err := e.Libvirt.DomainGetXMLDesc(domain, 0)
```
lukas016, line 126: don't forget implement some protection for multiple connection at the same time to the same machine. It isn't supported use case because we have only one session.

docs/development/setup.md
so-sahu: Added

docs/development/setup.md
so-sahu: Changes are made

provider/server/exec.go
so-sahu: I agree with you. Implemented a check both for nil device and empty console.

docs/development/setup.md
```diff
@@ -0,0 +1,143 @@
...
+### Setup `irictl-machine`
+
+1. **Clone ironcore repository**
+
+    ```bash
+    git clone git@github.com:ironcore-dev/ironcore.git
+    cd ironcore
+    ```
+
+2. **Build `irictl-machine`**
```
lukas016, line 37: you combine 2 way for number list. Explicit definition and auto increment. I recommend use only one way.

provider/server/exec.go
```diff
@@ -105,17 +106,25 @@ func (s *Server) ServeExec(w http.ResponseWriter, req *http.Request, token strin
...
+
+	// Check if a console is already active for this machine
+	if _, ok := activeConsoles.Load(machineID); ok {
+		return errors.New("operation failed: Active console session exists for this domain")
+	}
+
+	activeConsoles.Store(machineID, true)
+	defer activeConsoles.Delete(machineID)
+
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
so-sahu: done

docs/development/setup.md
```diff
@@ -0,0 +1,143 @@
...
+1. **Creating machine**
+
+    ```bash
+    irictl-machine --address=unix:<local-path-to-socket>/iri-machinebroker.sock create machine -f <path-to-machine-yaml>/iri-machine.yaml
+    ```
+
+    Sample `iri-machine.yaml`:
+
+    ```yaml
+        metadata:
```
lukas016, line 85: it looks little weird for me. Are you sure it is correct? I think metadata has to start directly under first `
Look on another code blocks, you don't have any extra whitespaces on beginning of line.

provider/server/exec.go
```diff
@@ -105,17 +104,22 @@ func (s *Server) ServeExec(w http.ResponseWriter, req *http.Request, token strin
 }
 
 func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser, _ remotecommand.TerminalSizeQueue) error {
-	var wg sync.WaitGroup
-
 	machineID := e.ExecRequest.MachineId
-	log := logr.FromContextOrDiscard(ctx).WithName(machineID)
+
+	// Check if a console is already active for this machine
+	if _, ok := activeConsoles.Load(machineID); ok {
```
lukas016, line 110: it doesn't look very safe. Combine Load and Store can be little unpredictable, if 2 goroutines run at the same time. I recommend use method LoadOrStore. It will be more safe

lukas016 (changes requested):
concurrency safety can be improved.

docs/development/setup.md
so-sahu: ok. I will set the orientations properly

provider/server/exec.go
so-sahu: You're right. Using Load followed by Store is unsafe in a concurrent environment. When two goroutines run simultaneously, both might pass the Load check and then both could try to Store the same machineID, causing a race condition. I'll switch to using the LoadOrStore method, which atomically loads the value for a key if it exists, or stores a new value if the key does not exist. This ensures that only one goroutine will successfully store the machineID as active. Thanks for pointing that out.

so-sahu: done

docs/development/setup.md
so-sahu: done

lukas016 (dismissed):
LGTM

lukas016:
Did you fix run of integration tests?

lukas016:
@so-sahu could you pls study this: https://github.com/mattn/go-tty/blob/master/tty_unix.go#L22
I am interesting why they open same file with 2 file descriptors. I think it can have good reason. You don't have to solve it in this PR but maybe we can create follow up issue for improvement.

provider/server/exec.go
```diff
@@ -124,21 +128,35 @@ func (e executorExec) Exec(ctx context.Context, in io.Reader, out io.WriteCloser
...
+		return fmt.Errorf("failed to unmarshal domain: %w", err)
 	}
 
-	cmd := exec.CommandContext(ctx, e.VirshExecutable, "-c", strings.TrimSpace(uri), "console", machineID)
+	if domainXML.Devices == nil || len(domainXML.Devices.Consoles) == 0 {
+		return errors.New("device console not set in machine domainXML")
+	}
+	ttyPath := domainXML.Devices.Consoles[0].TTY
 
-	f, err := pty.Start(cmd)
```
lukas016, line 134: why is permission 0? i could find any information about this setup. How is it working?

so-sahu: When opening an existing file for both reading and writing (O_RDWR) without the intention of creating a new file (O_CREATE not set), the permission parameter is disregarded. The file's current permissions are utilized, rendering the provided permission value ineffective.

Therefore, it is set to 0 in this case.

lukas016 (dismissed):
LGTM

provider/server/exec.go
```diff
@@ -26,13 +24,16 @@ import (
...
 	"k8s.io/client-go/tools/remotecommand"
+	"libvirt.org/go/libvirtxml"
 )
 
 const (
 	StreamCreationTimeout = 30 * time.Second
 	StreamIdleTimeout     = 2 * time.Minute
 )
 
+var activeConsoles sync.Map
```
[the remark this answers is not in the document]
lukasfrank: +1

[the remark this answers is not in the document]
so-sahu: As discussed with @lukas016 offline, it gets initialized everytime (https://github.com/ironcore-dev/libvirt-provider/blob/531b8fe9089a807c8075a523ab77c7ed55720d88/provider/server/exec.go#L86). Hence keeping it in that struct won't serve the purpose.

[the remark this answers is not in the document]
lukas016: @lukasfrank we discuss it offline and we are not very happy with current solution where streaming server internally calling iri server method. I think it has to be improve with refactor so i agree with current solution.
You can put it into Server structure but a lot of functionality are methods of Server so you won't have global variable but you will have variable accessible in all methods from Server object, what is for me extremely similar solution but i agree it will follow best practices.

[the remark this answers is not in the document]
lukasfrank: Since we anyhow have the `execRequestCache request.Cache[*iri.ExecRequest]` at the server which is used here, I would propose to attach the map to the server as well.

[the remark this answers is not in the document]
so-sahu: Done

lukas016 (approved):
LGTM

lukasfrank (approved):

PR #232, opened 2024-03-14
Move `DeleteMachine` call outside `Eventually` block in integration tests

### LINKED ISSUE #218: Refactor Machine Deletion Logic in Integration Tests
# Description

The current machine deletion logic in integration tests is causing multiple unnecessary calls to the libvirt-provider and flooding of logs in the console as seen [here](https://github.com/ironcore-dev/libvirt-provider/actions/runs/8136915210/job/22234141973). This is because the machine deletion is called inside the `Eventually` block, as seen in [this example](https://github.com/ironcore-dev/libvirt-provider/blob/206cd7787ff0cc0c8962ef9e4f21c45df85969e9/provider/server/machine_create_test.go#L59).

## Proposed Solution

Refactor the machine deletion logic in integration tests to avoid unnecessary calls and log flooding.

## Expected Behavior

Machine deletion logic should be optimized to avoid multiple unnecessary calls and log flooding.

## Additional Context

This enhancement will improve the efficiency and performance of integration tests by optimizing the machine deletion logic.

kasabe28:
Fixes #218

### DIFF
--- a/provider/server/exec_test.go
+++ b/provider/server/exec_test.go
@@ -37,18 +37,6 @@ var _ = Describe("Exec", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(createResp).NotTo(BeNil())
 
-		DeferCleanup(func(ctx SpecContext) {
-			Eventually(func(g Gomega) bool {
-				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
-				g.Expect(err).To(SatisfyAny(
-					BeNil(),
-					MatchError(ContainSubstring("NotFound")),
-				))
-				_, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
-				return libvirt.IsNotFound(err)
-			}).Should(BeTrue())
-		})
-
 		By("ensuring domain and domain XML is created for machine")
 		var domain libvirt.Domain
 		Eventually(func() error {
--- a/provider/server/exec_test.go
+++ b/provider/server/exec_test.go
@@ -37,6 +37,20 @@ var _ = Describe("Exec", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(createResp).NotTo(BeNil())
 
+		DeferCleanup(func(ctx SpecContext) {
+			Eventually(func(g Gomega) {
+				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+				g.Expect(err).To(SatisfyAny(
+					BeNil(),
+					MatchError(ContainSubstring("NotFound")),
+				))
+			}).Should(Succeed())
+			Eventually(func() bool {
+				_, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
+				return libvirt.IsNotFound(err)
+			}).Should(BeTrue())
+		})
+
 		By("ensuring domain and domain XML is created for machine")
 		var domain libvirt.Domain
 		Eventually(func() error {
--- a/provider/server/exec_test.go
+++ b/provider/server/exec_test.go
@@ -38,13 +38,13 @@ var _ = Describe("Exec", func() {
 		Expect(createResp).NotTo(BeNil())
 
 		DeferCleanup(func(ctx SpecContext) {
-			Eventually(func(g Gomega) {
+			Eventually(func() error {
 				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
-				g.Expect(err).To(SatisfyAny(
-					BeNil(),
-					MatchError(ContainSubstring("NotFound")),
-				))
-			}).Should(Succeed())
+				return err
+			}).Should(SatisfyAny(
+				BeNil(),
+				MatchError(ContainSubstring("NotFound")),
+			))
 			Eventually(func() bool {
 				_, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
 				return libvirt.IsNotFound(err)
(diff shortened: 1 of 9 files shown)


provider/server/machine_annotations_update_test.go
```diff
@@ -36,12 +36,9 @@ var _ = Describe("UpdateMachineAnnotations", func() {
 		Expect(createResp).NotTo(BeNil())
 
 		DeferCleanup(func(ctx SpecContext) {
+			_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+			Expect(err).To(BeNil())
```
lukas016, line 40: BDD requires using HaveOccurred or Succeed. more info here: https://onsi.github.io/gomega/#handling-errors

lukas016 (changes requested):

provider/server/machine_annotations_update_test.go
```diff
@@ -36,12 +36,9 @@ var _ = Describe("UpdateMachineAnnotations", func() {
 		Expect(createResp).NotTo(BeNil())
 
 		DeferCleanup(func(ctx SpecContext) {
+			_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+			Expect(err).To(Succeed())
```
so-sahu, line 40: I believe directly asserting the function call's result for success is more concise and easier to read.
eg: 
```suggestion
Expect(machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})).To(Succeed())

```

so-sahu (changes requested):
Function call's result can be directly checked for success.

provider/server/machine_annotations_update_test.go
kasabe28: `DeleteMachine` returns 2 values, so its not possible

provider/server/machine_create_test.go
```diff
@@ -55,12 +55,9 @@ var _ = Describe("CreateMachine", func() {
 		))
 
 		DeferCleanup(func(ctx SpecContext) {
+			_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
```
lukas016, line 58: i think tests can be unstable by this way. Secure solution is wrapping  DeleteMachine in own Eventually which will try call DeleteMachine maybe max 10second with 1second period, if DeleteMacine return error.

Root cause is place of calling function. Ideally DeferCleanup will left environment in the same state before tests were started.

lukas016 (changes requested):

provider/server/machine_create_test.go
lukas016: how were testes written never failed for DeleteMachine? i checked code and some edge cases can be occurred (unmarshall, unmarshall, write problem, wrong open file permissions, ...).  You can get error from machine store. They will be extremely rare and  many of them cannot be fix automatically probably (some can be) but implement secure solution isn't so hard and we can very easy improve stability of tests.
So pls add this additional Eventually blocks. thx

kasabe28: We have already secured solution by handling error from `DeleteMachine`. Adding it into Eventually block does not make any difference for `DeleteMachine` call, if it fail once then it will fail again also. That's why this issue has been raised to avoid multiple calls to `DeleteMachine`. If `DeleteMachine` fails somehow then genuinely there will be some error which Eventually() block will also not recover like you mentioned unmarshall, wrong file perms, these are non-recoverable by Eventually().

kasabe28: I had discussion with @so-sahu and we agreed that if this occurs in practical scenario then we will have separate PR for this.

lukas016: i will implement it.

provider/server/machine_annotations_update_test.go
```diff
@@ -36,12 +36,11 @@ var _ = Describe("UpdateMachineAnnotations", func() {
...
 		DeferCleanup(func(ctx SpecContext) {
-			Eventually(func(g Gomega) bool {
+			Eventually(func(g Gomega) {
 				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
-				g.Expect(err).To(SatisfyAny(
-					BeNil(),
-					MatchError(ContainSubstring("NotFound")),
-				))
+				g.Expect(err).Should(Succeed())
+			}).Should(Succeed())
```
so-sahu, line 42: It can be simplified:
```suggestion
Eventually(func() error {
	_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
	return err
}).Should(Succeed())
```
This eliminates the need for the additional Gomega argument as well. Please update other occurrences accordingly.

so-sahu (changes requested):
Eventually block for delete can be simplified.

provider/server/machine_annotations_update_test.go
lukas016: what is benefits? it looks more as cosmetic change. both solution are valid for gomega and this solution is more similar another eventually blocks.

provider/server/exec_test.go
```diff
@@ -38,12 +38,14 @@ var _ = Describe("Exec", func() {
 		Expect(createResp).NotTo(BeNil())
 
 		DeferCleanup(func(ctx SpecContext) {
-			Eventually(func(g Gomega) bool {
+			Eventually(func(g Gomega) {
```
afritzler, line 41: I don't think it is a good idea to call `DeleteMachine` in an `Eventually` block. We should cal the delete method once and then ensure that the domain has been removed.

afritzler (changes requested):

provider/server/machine_annotations_update_test.go
so-sahu: The functionality remains the same, but this approach offers advantages in terms of:

Readability: It explicitly conveys that the DeleteMachine call is expected to eventually succeed, eliminating the need for an additional Expect statement.

Clarity: Using Eventually(func() error { ... }).Should(Succeed()) directly signifies the expectation that an operation will ultimately succeed, avoiding the need for explicit error handling within the function.

However, this approach is optional and may vary based on preference.

lukas016: i will implement it.

provider/server/exec_test.go
lukas016: @afritzler do you understand why i want to have Eventually block for delete call? Can i close your comment?

lukas016:
test will be rewritten

PR #233, opened 2024-03-15
Removed `dd` dependency

### LINKED ISSUE #214: replace dd with golang implementation
# Summary

Libvirt-provider uses dd for create machine os image (https://github.com/ironcore-dev/libvirt-provider/blob/main/pkg/raw/raw_exec.go#L31)
Dependency to dd can be replaced with implementation written in golang and this dependency can be removed.

We have already solution in [virtlet](https://github.com/onmetal/virtlet/blob/main/raw/raw_exec.go#L58). It can be copy.

## Motivation

Remove dependencies for external tools.

so-sahu:
# Proposed Changes

Replacing the two dd commands used in raw_exec.go with two functions:

- a function designed to create an empty file at a specific seek position within the file system
- a function designed to copy the contents of one file to another while also providing the ability to specify a seek position within the destination file where the data will be written

Fixes https://github.com/ironcore-dev/libvirt-provider/issues/214

### DIFF
--- a/pkg/raw/raw_exec.go
+++ b/pkg/raw/raw_exec.go
@@ -5,42 +5,99 @@ package raw
 
 import (
 	"fmt"
-	"os/exec"
-	"strconv"
+	"io"
+	"os"
 
+	"github.com/go-logr/logr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-type Exec struct {
-}
+type Exec struct{}
+
+const filePerm = 0644
 
 func (Exec) Create(filename string, opts ...CreateOption) error {
 	o := &CreateOptions{}
 	o.ApplyOptions(opts)
-	var seek string
+	log := ctrl.Log
 
-	if o.SourceFile == "" && o.Size == nil {
-		return fmt.Errorf("must specify Size when creating without source file")
-	}
-
-	ofilename := "of=" + filename
-	if o.Size != nil {
-		seek = "seek=" + strconv.FormatInt(*o.Size, 10) // TODO: verify if the disk size is proper, else G as a suffix
-	}
 	if o.SourceFile == "" {
-		outMsgDD, err := exec.Command("dd", "if=/dev/zero", ofilename, "bs=1", "count=0", seek).Output()
+		if o.Size == nil {
+			return fmt.Errorf("must specify Size when creating without source file")
+		}
+		seek := *o.Size
+		// Position the file cursor one byte before the desired seek position to write a single byte,
+		// to ensure that data is written at the exact byte position specified by seek.
+		err := createEmptyFileWithSeek(log, filename, seek-1)
 		if err != nil {
-			return fmt.Errorf("failed creating the ephemeral disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating the empty ephemeral disk: %w", err)
 		}
 	} else {
-		ifilename := "if=" + o.SourceFile
-		outMsgDD, err := exec.Command("dd", ifilename, ofilename).Output()
+		err := copyFile(log, o.SourceFile, filename)
 		if err != nil {
-			return fmt.Errorf("failed creating the image for virtual disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating the image for virtual disk: %w", err)
 		}
 	}
 
 	return nil
+}
+
+func createEmptyFileWithSeek(log logr.Logger, filename string, seek int64) error {
+	log = log.WithName("createEmptyFileWithSeek")
+	dstFile, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
+			log.Error(destErr, "error closing file", "Path", dstFile)
+		}
+	}()
+
+	_, err = dstFile.Seek(seek, io.SeekStart)
+	if err != nil {
+		return err
+	}
+
+	_, err = dstFile.Write([]byte{0})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func copyFile(log logr.Logger, src, dst string) error {
+	log = log.WithName("copyFile")
+	srcFile, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		srcErr := srcFile.Close()
+		if srcErr != nil {
+			log.Error(srcErr, "error closing file", "Path", srcFile)
+		}
+	}()
+
+	dstFile, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
+			log.Error(destErr, "error closing file", "Path", dstFile)
+		}
+	}()
+
+	_, err = io.Copy(dstFile, srcFile)
+
+	return err
 }
 
 func init() {
--- a/pkg/raw/raw_exec.go
+++ b/pkg/raw/raw_exec.go
@@ -44,22 +44,22 @@ func (Exec) Create(filename string, opts ...CreateOption) error {
 }
 
 func createEmptyFileWithSeek(log logr.Logger, filename string, seek int64) error {
-	log = log.WithName("createEmptyFileWithSeek")
+	log = log.WithName("createEmptyFileWithSeek").WithValues("filename", filename)
 	dstFile, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed opening destination file: %w", err)
 	}
 
 	defer func() {
 		destErr := dstFile.Close()
 		if destErr != nil {
-			log.Error(destErr, "error closing file", "Path", dstFile)
+			log.Error(destErr, "error closing file")
 		}
 	}()
 
 	_, err = dstFile.Seek(seek, io.SeekStart)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed seeking destination file: %w", err)
 	}
 
 	_, err = dstFile.Write([]byte{0})
@@ -68,27 +68,27 @@ func createEmptyFileWithSeek(log logr.Logger, filename string, seek int64) error
 }
 
 func copyFile(log logr.Logger, src, dst string) error {
-	log = log.WithName("copyFile")
+	log = log.WithName("copyFile").WithValues("source", src, "destination", dst)
 	srcFile, err := os.Open(src)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed opening source file: %w", err)
 	}
 	defer func() {
 		srcErr := srcFile.Close()
 		if srcErr != nil {
-			log.Error(srcErr, "error closing file", "Path", srcFile)
+			log.Error(srcErr, "error closing source file")
 		}
 	}()
 
 	dstFile, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed opening destination file: %w", err)
 	}
 
 	defer func() {
 		destErr := dstFile.Close()
 		if destErr != nil {
-			log.Error(destErr, "error closing file", "Path", dstFile)
+			log.Error(destErr, "error closing destination file")
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,101 @@ package raw
...
+	"github.com/go-logr/logr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-type Exec struct {
-}
+type Exec struct{}
+
+const filePerm = 0644
```
lukas016, line 18: why 0644? i think 0660 or 0640 will be better. Little depends how vm will access to this file. So i mean libvirt-qemu user has to has access

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,101 @@ package raw
...
 
 func (Exec) Create(filename string, opts ...CreateOption) error {
 	o := &CreateOptions{}
 	o.ApplyOptions(opts)
-	var seek string
-
-	if o.SourceFile == "" && o.Size == nil {
-		return fmt.Errorf("must specify Size when creating without source file")
-	}
+	log := ctrl.Log
```
lukas016, line 23: optional: maybe we can create here logger with name raw-disk or something. It can  be parent logger for both createEmpty and copyFile loggers after. So we will see these log records have some bond.

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,101 @@ package raw
...
+		}
+	}()
+
+	_, err = dstFile.Seek(seek, io.SeekStart)
+	if err != nil {
+		return err
+	}
+
+	_, err = dstFile.Write([]byte{0})
+	if err != nil {
```
lukas016, line 66: optional: i recommend return err directly, i don't see big benefits in comparison on this level.

lukas016 (changes requested):

pkg/raw/raw_exec.go
so-sahu: changed it to 0640

pkg/raw/raw_exec.go
so-sahu: added WithName raw-disk

pkg/raw/raw_exec.go
so-sahu: done

pkg/raw/raw_exec.go
lukas016: are you sure?

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
+		if o.Size == nil {
+			return fmt.Errorf("must specify Size when creating without source file")
+		}
+		seek := *o.Size
+		// Position the file cursor one byte before the desired seek position to write a single byte,
+		// to ensure that data is written at the exact byte position specified by seek.
+		err := createEmptyFileWithSeek(log, filename, seek-1)
 		if err != nil {
-			return fmt.Errorf("failed creating the ephemeral disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating the empty ephemeral disk: %w", err)
```
lukas016, line 34: maybe add filename into logs. error on line 75 probably doesn't contain name of file what can be problem

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
+
+	dstFile, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
+			log.Error(destErr, "error closing file", "Path", dstFile)
```
lukas016, line 91: i think we are using lowercase for key (as first letter).

pkg/raw/raw_exec.go
so-sahu: The raw file is getting created with the user with which the controller is run. So, it looks like working.

pkg/raw/raw_exec.go
so-sahu: Added

pkg/raw/raw_exec.go
so-sahu: Removed path from log as it's included in top level using WithValues method.

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
-			return fmt.Errorf("failed creating the ephemeral disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating the empty ephemeral disk: %w", err)
 		}
 	} else {
-		ifilename := "if=" + o.SourceFile
-		outMsgDD, err := exec.Command("dd", ifilename, ofilename).Output()
+		err := copyFile(log, o.SourceFile, filename)
 		if err != nil {
-			return fmt.Errorf("failed creating the image for virtual disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating the image for virtual disk: %w", err)
```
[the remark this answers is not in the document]
so-sahu: Added

pkg/raw/raw_exec.go
lukas016: yes and this file has to be attach to vm, which need to do operation read and write. So if controller doesn't use libvirt-qemu user, probably you will end with permission error.

pkg/raw/raw_exec.go
lukas016: could you pls do refactor of error messages? Currently filename can be two times in error message. Little depends how error is created. So it doesn't look very consistent.

pkg/raw/raw_exec.go
so-sahu: As discussed offline, we have decided to provide users with more flexibility, allowing them to use different groups and users for each component. Therefore, we can set the permission to 0660 here.

pkg/raw/raw_exec.go
so-sahu: removed the additional filenames from the error messages of copyFile function. createEmptyFileWithSeek function does not contain additional filenames in error messages.

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
+	}()
+
+	dstFile, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	if err != nil {
+		return fmt.Errorf("failed opening destination file: %w", err)
+	}
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
```
lukasfrank, line 90:
```suggestion
		if err := dstFile.Close(); err != nil {
```

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
+
+func copyFile(log logr.Logger, src, dst string) error {
+	log = log.WithName("copyFile").WithValues("source", src, "destination", dst)
+	srcFile, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("failed opening source file: %w", err)
+	}
+	defer func() {
+		srcErr := srcFile.Close()
+		if srcErr != nil {
```
lukasfrank, line 78:
```suggestion
	
		if err := srcFile.Close(); err != nil {
```

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,98 @@ package raw
...
-			return fmt.Errorf("failed creating the image for virtual disk: %s, exit error %w", string(outMsgDD), err)
+			return fmt.Errorf("failed creating virtual disk image, source: %s, destination: %s: %w", o.SourceFile, filename, err)
 		}
 	}
 
 	return nil
 }
 
+func createEmptyFileWithSeek(log logr.Logger, filename string, seek int64) error {
+	log = log.WithName("createEmptyFileWithSeek").WithValues("filename", filename)
```
lukasfrank, line 47: Why do we need here a new logger if no logs except one error is being logged? Maybe just attach the filename to the logger while creating it in the `Create(filename string, opts ...CreateOption) error {` function.

lukasfrank (changes requested):

pkg/raw/raw_exec.go
so-sahu: done

pkg/raw/raw_exec.go
so-sahu: done

pkg/raw/raw_exec.go
so-sahu: Added the filename in the log itself.

As we are dealing with two files in the overall program i.e source file and destination file, attaching the filename to the logger while creating it in the `Create(filename string, opts ...CreateOption) error {` function, could lead to confusion in debugging and maintenance.

Keeping the WithName method, as having it for each component will help in providing clear and identifiable log messages, which is generally preferred for better code readability and maintainability.

pkg/raw/raw_exec.go
lukas016: don't use shadowing you can very easy did mistake in code because one stupid character : can completely change logic.

pkg/raw/raw_exec.go
lukas016: logger is primary for additional logs in Close functions. These errors aren't fatal but they can be problematic for long run apps. So we log them minimally.

lukasfrank: logging them is totally fine. but it dosen't really make sense to create a new logger for every function. This code belongs to the `Create` function and with a meaningful error the logger passed to `createEmptyFileWithSeek(log logr.Logger` can easily used.

lukas016: sorry i didn't understand your post properly. Filename can be added in Create function. @so-sahu we don't need add target filename because logic is hardcoded here, so i don't see big benefits with logging of destination file. But i recommend keep name of function somewhere in logs for better debugging.

pkg/raw/raw_exec.go
so-sahu: Yes, I agree and had implemented this way only. However, @lukasfrank wanted to have this sway. Let's finalize it offline.

pkg/raw/raw_exec.go
so-sahu: Implemented the change, as discussed offline.

pkg/raw/raw_exec.go
so-sahu: As per our offline discussion, this aligns with our current development standards, so we'll proceed with this approach.

lukasfrank (changes requested):

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,93 @@ package raw
...
+		return fmt.Errorf("failed opening destination file: %w", err)
+	}
+
+	defer func() {
+		if err := dstFile.Close(); err != nil {
+			log.Error(err, "error closing destination file in copyFile")
+		}
+	}()
+
+	_, err = io.Copy(dstFile, srcFile)
```
lukasfrank, line 90: Can you please extend here the error?

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,93 @@ package raw
...
+			log.Error(err, "error closing file in createEmptyFileWithSeek")
+		}
+	}()
+
+	_, err = dstFile.Seek(seek, io.SeekStart)
+	if err != nil {
+		return fmt.Errorf("failed seeking destination file: %w", err)
+	}
+
+	_, err = dstFile.Write([]byte{0})
```
lukasfrank, line 63: Can you please extend here the error?

so-sahu: done

pkg/raw/raw_exec.go
so-sahu: done

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,94 @@ package raw
...
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
+			log.Error(destErr, "error closing file in createEmptyFileWithSeek")
+		}
+	}()
+
+	_, err = dstFile.Seek(seek, io.SeekStart)
+	if err != nil {
```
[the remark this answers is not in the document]
so-sahu: done

pkg/raw/raw_exec.go
```diff
@@ -5,44 +5,94 @@ package raw
...
 
+func createEmptyFileWithSeek(log logr.Logger, filename string, seek int64) error {
+	dstFile, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, filePerm)
+	if err != nil {
+		return fmt.Errorf("failed opening destination file: %w", err)
+	}
+
+	defer func() {
+		destErr := dstFile.Close()
+		if destErr != nil {
```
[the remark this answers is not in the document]
so-sahu: done

lukas016 (approved):
LGTM

lukasfrank (approved):

PR #243, opened 2024-03-19
Filter machine lifecycle events

### LINKED ISSUE #226: Reconciliation of Unmanaged Machines in `Libvirt-Provider`
**Describe the bug**
After integrating support for libvirt events (issue #177), it has been observed that machines not managed by the libvirt provider (i.e., not present in the store) are also triggering requeue requests due to their lifecycle events. The libvirt provider should ideally only reconcile machines that it manages, i.e., machines present in the store.

**To Reproduce**
- Create a machine using the virsh command.
- Start the libvirt provider.
- Destroy the machine created in step 1.
- Observe the requeue request in the libvirt provider console for the machine that was destroyed.

**Expected behavior**
The libvirt provider should only reconcile machines that it manages, i.e., machines present in the store.

**Proposed Solution**
Implement a check in the libvirt provider to ensure that only machines present in the store trigger requeue requests and undergo reconciliation. Any events related to unmanaged machines should be ignored.

**Additional context**
This issue is critical for maintaining the efficiency and performance of the libvirt provider in environments with a mix of managed and unmanaged machines.

lukasfrank:
# Proposed Changes

- Only enqueue events from `Machine`s managed by `libvirt-provider`

Fixes https://github.com/ironcore-dev/libvirt-provider/issues/226

### DIFF
--- a/pkg/controllers/machine_controller.go
+++ b/pkg/controllers/machine_controller.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
-	"strconv"
 	"sync"
 	"time"
 
@@ -260,7 +259,7 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
 }
 
 func (r *MachineReconciler) startEnqueueMachineByLibvirtEvent(ctx context.Context, log logr.Logger) {
-	eventChan, err := r.libvirt.LifecycleEvents(ctx)
+	lifecycleEvents, err := r.libvirt.LifecycleEvents(ctx)
 	if err != nil {
 		log.Error(err, "failed to subscribe to libvirt lifecycle events")
 		return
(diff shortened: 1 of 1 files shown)


pkg/controllers/machine_controller.go
```diff
@@ -270,14 +269,24 @@ func (r *MachineReconciler) startEnqueueMachineByLibvirtEvent(ctx context.Contex
...
+			if err != nil {
+				if errors.Is(err, store.ErrNotFound) {
+					log.V(2).Info("Skipped: not managed by libvirt-provider", "machineID", machine.ID)
+					continue
+				}
+				log.Error(err, "failed to fetch machine from store")
+				continue
+			}
+
+			log.V(1).Info("requeue machine", "machineID", machine.ID, "eventID", evt.Event)
```
lukas016, line 288: eventID isn't very correct. It is more subevent ID in context.
```suggestion
			log.V(1).Info("requeue machine", "machineID", machine.ID, "lifecycleEventID", evt.Event)
```

lukas016 (changes requested):
logging need to be improve

pkg/controllers/machine_controller.go
lukasfrank: fixed

lukas016 (approved):

Issue #256, opened 2024-03-27
Machine classes not linked to created machine pool during development deployment


so-sahu:
**Description:**
During development deployment, the machine classes are not getting linked to the created machine pool. After conducting Root Cause Analysis, it was identified that in the [kustomization.yaml](https://github.com/ironcore-dev/libvirt-provider/blob/0b16082f331c888e6f4bf1826fb9b71b9c142f7f/config/development/kustomization.yaml#L4) file, a `namePrefix` is set to `libvirt-provider-dev-`. To resolve this issue, the names of machine classes in the [machineclasses.json](https://github.com/ironcore-dev/libvirt-provider/blob/main/config/development/machineclasses.json) file should be prefixed with `libvirt-provider-dev-`.

**To Reproduce**
- Apply the development kustomization config.
- List the machinePool and observe the linked machineClass.

**Expected behavior**
The machineClasses should get linked with the machinePool.

lukasfrank:
The class including all properties needs to match. This is an expected behaviour.

PR #257, opened 2024-03-27
Rectifying MachineClass Names in Development Deployment

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/256

### DIFF
--- a/config/development/machineclasses.json
+++ b/config/development/machineclasses.json
@@ -1,20 +1,20 @@
 [
   {
-    "name": "extra-huge",
+    "name": "libvirt-provider-dev-extra-huge",
     "capabilities": {
       "cpu_millis": 2000,
       "memory_bytes": 16106127361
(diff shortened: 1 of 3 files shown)


config/development/machineclasses.json
```diff
@@ -1,20 +1,20 @@
 [
   {
-    "name": "extra-huge",
+    "name": "libvirt-provider-dev-extra-huge",
```
lukas016, line 3: why? was it problem?

so-sahu: Yes, according to the current configuration, applying the customization results in all resources being created and the DaemonSet coming to a running state. However, when querying the machine pool, the machine classes are blank. This issue prevents the machine from being reconciled in the pod.

lukas016 (approved):
LGTM

Issue #167, opened 2024-02-08
overcommit vcpu for scheduling


lukas016:
# Summary

We need to report different count of vcpu cores for allocation as we really have count of available physical cores.
Main reason is better usage of hw, when cpu cores don't have to be used on 100%.

This feature has to be optional and configurable with argument.
We can increase and decrease count of vcpu.

## Basic example

Default behavior count of vcpu is the same as count of physical cores:
--resource-manager-overcommit-vcpu 1.0

Decrease count of vcpu to half count of physical cores:
--resource-manager-overcommit-vcpu 0.5

Increase count of vcpu to double count of physical cores:
--resource-manager-overcommit-vcpu 2.0

## Motivation

Better using of cpus.

lukas016:
merged

PR #264, opened 2024-04-02
Overcommit VCPU for Scheduling

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/167

### DIFF
--- a/pkg/resources/sources/cpu.go
+++ b/pkg/resources/sources/cpu.go
@@ -5,6 +5,7 @@ package sources
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/shirou/gopsutil/v3/cpu"
@@ -13,10 +14,12 @@ import (
 	core "github.com/ironcore-dev/ironcore/api/core/v1alpha1"
 )
 
-type CPU struct{}
+type CPU struct {
+	OvercommitVCPU float64
+}
 
-func NewSourceCPU() *CPU {
-	return &CPU{}
+func NewSourceCPU(overCommitVCPU float64) *CPU {
+	return &CPU{OvercommitVCPU: overCommitVCPU}
 }
 
 func (c *CPU) GetTotalResources(ctx context.Context) (core.ResourceList, error) {
@@ -36,8 +39,8 @@ func (c *CPU) GetTotalResources(ctx context.Context) (core.ResourceList, error) 
 	return resources, nil
 }
 
-func (c *CPU) GetName() string {
-	return "cpu"
+func (c *CPU) GetName() core.ResourceName {
+	return core.ResourceCPU
 }
 
 // Modify rounding up cpu to total cores
@@ -49,6 +52,28 @@ func (c *CPU) Modify(resources core.ResourceList) error {
 
 	cpu.RoundUp(resource.Kilo)
 	resources[core.ResourceCPU] = cpu
+
+	return nil
+}
+
+func (c *CPU) TuneTotalResources(resources core.ResourceList) error {
+	if c.OvercommitVCPU == 0 || c.OvercommitVCPU < 0 {
+		return errors.New("overcommitVCPU cannot be zero or negative")
+	}
+
+	cpuQuantity, ok := resources[core.ResourceCPU]
+	if !ok {
+		return errors.New("failed to get CPU resource from resource manager")
+	}
+
+	totalCPU, ok := cpuQuantity.AsInt64()
+	if !ok {
+		return errors.New("failed to convert CPU quantity to int64")
+	}
+
+	totalCPUFloat := float64(totalCPU) * c.OvercommitVCPU
+
+	resources[core.ResourceCPU] = *resource.NewQuantity(int64(totalCPUFloat), resource.DecimalSI)
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


pkg/resources/sources/cpu.go
```diff
@@ -31,13 +37,16 @@ func (c *CPU) GetTotalResources(ctx context.Context) (core.ResourceList, error)
...
-	resources := core.ResourceList{core.ResourceCPU: *resource.NewScaledQuantity(hostCPUSum, resource.Kilo)}
+	cpuQuantity := float64(hostCPUSum) * c.OvercommitVCPU
+	resources := core.ResourceList{
+		core.ResourceCPU: *resource.NewQuantity(int64(cpuQuantity*1000), resource.DecimalSI),
+	}
 
 	return resources, nil
 }
 
-func (c *CPU) GetName() string {
```
lukas016, line 39: why are you mixing source name with resource name?

pkg/resources/sources/cpu.go
```diff
@@ -13,13 +14,18 @@ import (
...
+}
 
-func NewSourceCPU() *CPU {
-	return &CPU{}
+func NewSourceCPU(options Options) *CPU {
+	return &CPU{OvercommitVCPU: options.OvercommitVCPU}
 }
 
 func (c *CPU) GetTotalResources(ctx context.Context) (core.ResourceList, error) {
+	if c.OvercommitVCPU == 0 || c.OvercommitVCPU < 0 {
```
lukas016, line 26: i recommend use https://pkg.go.dev/math#Signbit instead of < 0

pkg/resources/sources/cpu.go
so-sahu: ok, I will create separate constants for the source names.

so-sahu: However, we already have the source name associated with the resource name: [memory.go](https://github.com/ironcore-dev/libvirt-provider/blob/b1d44ebf879bfb8ea48a54a84198d8e827376e19/pkg/resources/sources/memory.go#L17).

lukas016: no, it is resource name for new resource, but in GetName() i returned hugepages strings

pkg/resources/sources/cpu.go
```diff
@@ -31,13 +37,16 @@ func (c *CPU) GetTotalResources(ctx context.Context) (core.ResourceList, error)
 		hostCPUSum += int64(v.Cores)
 	}
 
-	resources := core.ResourceList{core.ResourceCPU: *resource.NewScaledQuantity(hostCPUSum, resource.Kilo)}
+	cpuQuantity := float64(hostCPUSum) * c.OvercommitVCPU
+	resources := core.ResourceList{
```
lukas016, line 41: why did you change this operation and you didn't used NewScaledQuantity?

so-sahu: The change was made because NewScaledQuantity doesn't handle floating-point scaling well. NewScaledQuantity expects an integer as the first argument and a resource.Scale as the second argument. Since c.OvercommitVCPU can be a float, we needed to perform the scaling manually by converting the result to an integer. This manual scaling ensures that fractional values like 1.5 or 0.5 are correctly converted to their corresponding values in kilo. Using NewScaledQuantity directly wouldn't have allowed us to handle such fractional values correctly.

pkg/resources/sources/cpu.go
so-sahu: Got it. I'll create a separate constant for the source name and use that instead of hardcoding strings in multiple places to ensure it can be managed from a single location.

pkg/resources/sources/cpu.go
lukas016: i don't think it will safe too. comparing float to specific number isn't (minimally in c language) safe operation.
so maybe something as c.OvercommitVCPU < 0.0001 can be safer.

pkg/resources/sources/cpu.go
lukas016: but you don't use float and you retype it to int64, so i am confuse.

so-sahu: The reason for manually scaling the CPU quantity and converting it to an integer (int64) is to ensure that fractional values, such as 1.5 or 0.5, are correctly converted to their corresponding values in kilo. Even though the c.OvercommitVCPU value may be a float, the final result needs to be an integer representing the CPU quantity in kilo.

The NewScaledQuantity function expects an integer as its first argument, which is why we need to perform the scaling manually. By converting the result to an integer, we ensure that the CPU quantity is correctly represented in kilo, even if the c.OvercommitVCPU value is a float.

Using NewScaledQuantity directly wouldn't allow us to handle such fractional values correctly, as it expects an integer value for scaling. Therefore, the manual scaling and conversion to an integer are necessary in this case.

pkg/resources/sources/cpu.go
so-sahu: Using `epsilon := 1e-9` for the comparison. It will solve the issue.

pkg/resources/sources/cpu.go
so-sahu: As discussed offline, using `NewScaledQuantity`

lukas016 (approved):

PR #272, opened 2024-04-08
Restructuring Resource Manager

so-sahu:
# Proposed Changes

Restructuring resource manager

### DIFF
--- a/pkg/resources/manager/manager.go
+++ b/pkg/resources/manager/manager.go
@@ -13,11 +13,11 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/golang-collections/collections/set"
 	core "github.com/ironcore-dev/ironcore/api/core/v1alpha1"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
 	"github.com/ironcore-dev/libvirt-provider/pkg/api"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/go-logr/logr"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -166,7 +166,7 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
 	r.ctx = ctx
 
 	// Initialize all sources and check for common resources
-	var initializedResources *set.Set
+	var initializedResources sets.Set[core.ResourceName]
 	for _, s := range r.sources {
 		resources, err := s.Init(r.ctx)
 		if err != nil {
@@ -175,14 +175,15 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
 
 		if initializedResources == nil {
 			initializedResources = resources
-		} else {
-			// Check for common resources with previously initialized sources
-			if initializedResources.Intersection(resources).Len() > 0 {
-				return fmt.Errorf("%w: source `%s` has resource(s): `%s`, common with previously initialized resources: `%s`", ErrCommonResources, s.GetName(), getElementsFromSet(resources), getElementsFromSet(initializedResources))
-			}
-
-			initializedResources = initializedResources.Union(resources)
+			continue
 		}
+		// Check for common resources with previously initialized sources
+		commonSources := initializedResources.Intersection(resources)
+		if commonSources.Len() > 0 {
+			return fmt.Errorf("%w: source '%s' has resource(s): '%s', that are already initialized", ErrCommonResources, s.GetName(), getElementsFromSet(commonSources))
+		}
+
+		initializedResources = initializedResources.Union(resources)
 	}
 
 	r.log.Info("Initialized resources: " + r.convertResourcesToString(r.getAvailableResources()))
@@ -453,11 +454,11 @@ func mergeResourceLists(dst, src core.ResourceList) {
 }
 
 // getElementsFromSet returns a string representation of the set's contents
-func getElementsFromSet(s *set.Set) string {
-	var elements []string
-	s.Do(func(elem interface{}) {
-		elements = append(elements, fmt.Sprintf("%v", elem))
-	})
+func getElementsFromSet(s sets.Set[core.ResourceName]) string {
+	elements := make([]string, 0, s.Len())
+	for _, elem := range s.UnsortedList() {
+		elements = append(elements, string(elem))
+	}
 	return strings.Join(elements, ", ")
 }
 
--- a/pkg/resources/manager/manager.go
+++ b/pkg/resources/manager/manager.go
@@ -15,6 +15,7 @@ import (
 	core "github.com/ironcore-dev/ironcore/api/core/v1alpha1"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
 	"github.com/ironcore-dev/libvirt-provider/pkg/api"
+	"github.com/ironcore-dev/libvirt-provider/pkg/resources/sources"
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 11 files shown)


pkg/resources/manager/manager.go
```diff
@@ -32,14 +33,13 @@ var (
...
-	ErrResourceAlreadyRegistered = errors.New("resource is already registered")
-	ErrResourceNegativeQuantity  = errors.New("resource quantity is negative")
+	ErrResourceNotAvailable     = errors.New("not enough available resources")
+	ErrResourceUnsupported      = errors.New("resource isn't supported")
+	ErrResourceNegativeQuantity = errors.New("resource quantity is negative")
 
 	ErrMachineClassMissing = errors.New("machine class is missing")
 
-	ErrMachineHasAllocatedResources = errors.New("machine has already allocated resources")
+	ErrIncompatibleSources = errors.New("sources can not contain both memory and hugepages")
```
lukas016, line 42: message isn't enough general.

pkg/resources/sources/dummy.go
```diff
@@ -35,3 +35,19 @@ func (d *Dummy) GetName() string {
 func (d *Dummy) Modify(_ core.ResourceList) error {
 	return nil
 }
+
+func (d *Dummy) Allocate(requiredResources core.ResourceList) core.ResourceList {
+	return nil
```
lukas016, line 40: why?

pkg/resources/sources/dummy.go
```diff
@@ -35,3 +35,19 @@ func (d *Dummy) GetName() string {
...
+}
+
+func (d *Dummy) Deallocate(requiredResources core.ResourceList) []core.ResourceName {
+	return nil
```
lukas016, line 44: why?

pkg/resources/sources/dummy.go
```diff
@@ -35,3 +35,19 @@ func (d *Dummy) GetName() string {
...
+}
+
+func (d *Dummy) GetAvailableResource() core.ResourceList {
+	return core.ResourceList{core.ResourceCPU: *d.totalResources.CPU(), core.ResourceMemory: *d.totalResources.Memory()}
```
lukas016, line 48: dummy has to know manage any type of resource

pkg/resources/manager/manager.go
```diff
@@ -239,11 +218,24 @@ func (r *resourceManager) allocate(machine *api.Machine, requiredResources core.
...
 	}
 
-	newAvailableResources, err := r.preallocateAvailableResources(requiredResources)
-	if err != nil {
-		return err
+	totalAllocatedRes := core.ResourceList{}
+	for _, s := range r.sources {
+		allocatedRes := s.Allocate(requiredResources)
+
+		sourceResourceList := s.GetAvailableResource()
```
lukas016, line 225: this logic is wrong, it cannot work properly.

pkg/resources/manager/manager.go
```diff
@@ -239,11 +218,24 @@ func (r *resourceManager) allocate(machine *api.Machine, requiredResources core.
...
+		sourceResourceQuantity := sourceResourceList[core.ResourceName(s.GetName())]
+		if sourceResourceQuantity.Sign() == -1 {
+			_ = s.Deallocate(requiredResources)
+			return ErrResourceNotAvailable
+		} else {
+			for k, v := range allocatedRes {
```
lukas016, line 231: create additional function for that

pkg/resources/manager/manager.go
```diff
@@ -472,6 +431,18 @@ func (r *resourceManager) getMachineClassAvailibilityAsString() string {
 	return removeSeparatorFromEnd(result)
 }
 
+func (r *resourceManager) getResourceList() core.ResourceList {
```
lukas016, line 434: name of function is very general. It will always return available resources.

pkg/resources/manager/manager.go
```diff
@@ -296,31 +265,20 @@ func (r *resourceManager) deallocate(machine *api.Machine, deallocateResources c
...
-		newResources[key] = *available
-	}
-
 	if r.numaScheduler != nil {
 		err = r.numaScheduler.Unpin(machine)
 		if err != nil {
 			return err
 		}
 	}
+	for _, s := range r.sources {
```
lukas016, line 274: this cannot work.

lukas016 (changes requested):

pkg/resources/manager/manager.go
lukas016: my fault, it will work

pkg/resources/manager/manager.go
so-sahu: added generalised message.

pkg/resources/manager/manager.go
so-sahu: I was mistaken. It is only used here, so a separate function is not necessary.

pkg/resources/manager/manager.go
so-sahu: changed

pkg/resources/manager/manager.go
```diff
@@ -202,24 +171,34 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
...
-		return err
+	for _, s := range r.sources {
+		err := s.Init(r.ctx)
+		if err != nil {
+			return err
+		}
 	}
 
-	//TODO implement "limiters"
+	r.log.Info("Total resources: " + r.convertResourcesToString(r.getResourceList()))
```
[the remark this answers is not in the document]
so-sahu: done

pkg/resources/manager/manager.go
```diff
@@ -193,6 +151,10 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
 	r.mx.Lock()
 	defer r.mx.Unlock()
 
+	if len(r.sources) == 0 {
```
lukas016, line 154: cannot we move it into initialize function? do we want to support resource manager without resources?

pkg/resources/manager/manager.go
```diff
@@ -472,6 +436,18 @@ func (r *resourceManager) getMachineClassAvailibilityAsString() string {
 	return removeSeparatorFromEnd(result)
 }
 
+func (r *resourceManager) getAvailableResources() core.ResourceList {
+	resourceList := make(core.ResourceList)
+
+	for _, s := range r.sources {
+		list := s.GetAvailableResource().DeepCopy()
```
lukas016, line 443: i recommend break reference directly in sources

pkg/resources/manager/manager.go
```diff
@@ -202,24 +164,47 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
...
+	var initializedResources *set.Set
+	for _, s := range r.sources {
+		resources, err := s.Init(r.ctx)
+		if err != nil {
+			return err
+		}
+
+		if initializedResources == nil {
+			initializedResources = resources
+		} else {
```
lukas016, line 177: this else block isn't needed

pkg/resources/manager/manager.go
```diff
@@ -202,24 +164,47 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
...
+		}
 	}
 
-	//TODO implement "limiters"
+	r.log.Info("Initialized resources: " + r.convertResourcesToString(r.getAvailableResources()))
 
-	err = r.calculateAvailableResources(machines)
-	if err != nil {
-		return err
+	for _, machine := range machines {
```
lukas016, line 189: add here maybe comments what it is doing or put it into function.

pkg/resources/manager/manager.go
so-sahu: It is needed, else in the first iteration, it will have all the common resources

pkg/resources/manager/manager.go
so-sahu: Comment added.

pkg/resources/manager/manager.go
```diff
@@ -202,24 +164,47 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
...
+		if err != nil {
+			return err
+		}
+
+		if initializedResources == nil {
+			initializedResources = resources
+		} else {
+			// Check for common resources with previously initialized sources
+			if initializedResources.Intersection(resources).Len() > 0 {
+				return ErrCommonResources
```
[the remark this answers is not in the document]
so-sahu: Added a more descriptive error message.

pkg/resources/manager/manager.go
so-sahu: Done

lukas016 (approved):

Issue #166, opened 2024-02-08
reserve hugepages / memory which cannot be used vm


lukas016:
# Summary

We need to reserve memory bytes and count of hugepages which cannot be allocated for vm.
Main reason is calculation of capacity based on total memory/total hugepages size. This capacity isn't fully available because memory is consume by k3s, os and other apps. Minimally we need reserve 2 hugepages for dp-service for proper working.

It has to be configurable with arguments and default value can be 0.

## Basic example

--resource-manager-reserve-memory-size=8Gi
--resource-manager-reserve-hugepages-count=2

## Motivation

Improve scheduling and minimize side effects to other apps.

lukas016:
merged

PR #277, opened 2024-04-11
Reserving Hugepages and Memory

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/166

### DIFF
--- a/pkg/resources/sources/memory.go
+++ b/pkg/resources/sources/memory.go
@@ -179,8 +179,8 @@ func (m *Hugepages) GetAvailableResource() core.ResourceList {
 }
 
 func calculateAvailableMemory(totalMemory, reservedMemory MemorySize) (*resource.Quantity, error) {
-	if reservedMemory >= totalMemory {
-		return nil, fmt.Errorf("reservedMemorySize cannot be greater than equal to totalMemory: %v", resource.NewQuantity(int64(totalMemory), resource.BinarySI))
+	if reservedMemory > totalMemory {
+		return nil, fmt.Errorf("reservedMemorySize cannot be greater than totalMemory: %v", resource.NewQuantity(int64(totalMemory), resource.BinarySI))
 	}
 	availableMemoryUint := MemorySize(totalMemory) - reservedMemory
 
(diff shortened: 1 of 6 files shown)


pkg/resources/sources/options.go
```diff
@@ -3,7 +3,55 @@
...
+	Sources            []string
+	OvercommitVCPU     float64
+	ReservedMemorySize MemorySize
+	BlockedHugepages   uint64
+}
+
+// MemorySize is a custom type to handle memory sizes in human-readable format.
+type MemorySize uint64
+
+func (m *MemorySize) String() string {
```
lukas016, line 21: here you can create copy (m MemorySize) pointer and uint64 have same size so you don't have any benefits with pointer.

pkg/resources/sources/options.go
```diff
@@ -3,7 +3,55 @@
...
-	Sources        []string
-	OvercommitVCPU float64
+	Sources            []string
+	OvercommitVCPU     float64
+	ReservedMemorySize MemorySize
+	BlockedHugepages   uint64
+}
+
+// MemorySize is a custom type to handle memory sizes in human-readable format.
+type MemorySize uint64
```
lukas016, line 19: why? cannot you use any existing library e.g.: https://github.com/docker/go-units/blob/master/size.go

pkg/resources/sources/memory.go
```diff
@@ -116,8 +123,17 @@ func (m *Hugepages) Init(ctx context.Context) (sets.Set[core.ResourceName], erro
 	m.pageSize = hostMem.HugePageSize
 	m.pageCount = hostMem.HugePagesTotal
 
-	m.availableMemory = resource.NewQuantity(int64(m.pageSize*m.pageCount), resource.BinarySI)
-	m.availableHugePages = resource.NewQuantity(int64(m.pageCount), resource.DecimalSI)
+	availableMemory, err := calculateAvailableMemory(MemorySize(m.pageSize*m.pageCount), m.reservedMemorySize)
```
lukas016, line 126: it doesn't look good. are you sure it is working? because memory capacity is calculated from hugepages and here you calculate reservedmemory independent.

it looks confusing.

so-sahu: memory calculation changed as discussed.

pkg/resources/sources/options.go
so-sahu: Used

pkg/resources/sources/options.go
so-sahu: Done

lukas016 (approved):

Issue #172, opened 2024-02-09
limit count of vms which can be created


lukas016:
# Summary

We need to implement support for limitation of count of vms which can be created/manage by libvirt-provider. Ideal place for the logic will be resource manager.
Main reason is limiting count of nic which we can assign to machine (128 nics). Implementing support for nics in resource management is more complex issue, so we need simpler solution now.
This feature can be used as general/temporary solution for another edge cases which can show up in future.
Example another edge cases:
- performance issue, if you have more as 20 machine

## Basic example

You can create max 20 vms
--resource-manager-vm-limit=20

Feature is disabled
--resource-manager-vm-limit=0

It cannot be negative number.

## Motivation

Create simple solution for limit count of vms which can be created because we have limited count of nics. It can be as general solution for another edge cases.

lukas016:
merged

PR #279, opened 2024-04-16
limiting Count of VMs

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/172

### DIFF
--- a/pkg/resources/manager/functions.go
+++ b/pkg/resources/manager/functions.go
@@ -28,10 +28,6 @@ func Allocate(machine *api.Machine, requiredResources core.ResourceList) error {
 		return ErrResourcesEmpty
 	}
 
-	if mng.vmCountLimit != 0 && mng.vmCount >= mng.vmCountLimit {
-		return ErrVMLimitReached
-	}
-
 	return mng.allocate(machine, requiredResources)
 }
 
@@ -61,8 +57,8 @@ func SetMachineClasses(classes []*iri.MachineClass) error {
 }
 
 // SetVMLimit just registers maximum limit for VMs
-func SetVMLimit(vmCountLimit int) error {
-	return mng.setVMLimit(vmCountLimit)
+func SetVMLimit(maxVMsLimit uint64) error {
+	return mng.setVMLimit(maxVMsLimit)
 }
 
 // GetMachineClassStatus return status of machineclasses with current quantity
--- a/pkg/resources/manager/functions.go
+++ b/pkg/resources/manager/functions.go
@@ -27,6 +27,11 @@ func Allocate(machine *api.Machine, requiredResources core.ResourceList) error {
 	if len(requiredResources) == 0 {
 		return ErrResourcesEmpty
 	}
+
+	if mng.vmCountLimit != 0 && mng.vmCount >= mng.vmCountLimit {
+		return ErrVMLimitReached
+	}
+
 	return mng.allocate(machine, requiredResources)
 }
 
@@ -53,6 +58,11 @@ func SetLogger(logger logr.Logger) error {
 // SetMachineClasses just registers supported machineclasses
 func SetMachineClasses(classes []*iri.MachineClass) error {
 	return mng.setMachineClasses(classes)
+}
+
+// SetVMLimit just registers maximum limit for VMs
+func SetVMLimit(vmCountLimit int) error {
+	return mng.setVMLimit(vmCountLimit)
 }
 
 // GetMachineClassStatus return status of machineclasses with current quantity
--- a/pkg/resources/manager/manager.go
+++ b/pkg/resources/manager/manager.go
@@ -79,11 +79,11 @@ type resourceManager struct {
 	// and it serves as protection for calling function before initialization.
 	operationError error
 
-	// current number of VMs on host
-	vmCount int
+	// availableVMSlots signifies the number of VMs that can still be created on the host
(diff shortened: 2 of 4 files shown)


provider/cmd/app/app.go
```diff
@@ -153,6 +153,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.Float64Var(&o.ResourceManagerOptions.OvercommitVCPU, "resource-manager-overcommit-vcpu", 1.0, "Sets the overcommit ratio for vCPUs, enabling higher VM density per CPU core.")
 	fs.Uint64Var(&o.ResourceManagerOptions.BlockedHugepages, "resource-manager-blocked-hugepages", 0, "Count of hugepages which aren't use for vms. Effective only if hugepages source is set")
 	fs.Var(&o.ResourceManagerOptions.ReservedMemorySize, "resource-manager-reserved-memory-size", "Size of memory which aren't use for vms in human-readable format. Effective only if memory source is set")
+	fs.Uint64Var(&o.ResourceManagerOptions.VMLimit, "resource-manager-vm-limit", 0, "Maximum number of the VMs to be created on the host")
```
lukas016, line 156: i think 0 can be valid values for some edge cases. So maybe it will be better use negative numbers for disable this option.

pkg/resources/sources/options.go
```diff
@@ -12,6 +12,7 @@ type Options struct {
 	OvercommitVCPU     float64
 	ReservedMemorySize MemorySize
 	BlockedHugepages   uint64
+	VMLimit            uint64
```
lukas016, line 15: recommend use the type as it is use later.

provider/cmd/app/app.go
```diff
@@ -556,6 +557,11 @@ func initResourceManager(ctx context.Context, opts sources.Options, machineStore
 		return err
 	}
 
+	err = manager.SetVMLimit(int(opts.VMLimit))
```
lukas016, line 560: this retype can change value it isn't secure/

pkg/resources/manager/functions.go
```diff
@@ -55,6 +60,11 @@ func SetMachineClasses(classes []*iri.MachineClass) error {
 	return mng.setMachineClasses(classes)
 }
 
+// SetVMLimit just registers maximum limit for VMs
+func SetVMLimit(vmCountLimit int) error {
```
lukas016, line 64: i am not big fan of type without size. Is possible use int32?

pkg/resources/manager/functions.go
```diff
@@ -27,6 +27,11 @@ func Allocate(machine *api.Machine, requiredResources core.ResourceList) error {
 	if len(requiredResources) == 0 {
 		return ErrResourcesEmpty
 	}
+
+	if mng.vmCountLimit != 0 && mng.vmCount >= mng.vmCountLimit {
```
lukas016, line 31: use it here can be dangerous because vmCount can be change with separate goroutine

pkg/resources/manager/manager.go
```diff
@@ -165,6 +183,12 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
 
 	r.ctx = ctx
 
+	totalExistingVMCount := len(machines)
+	if r.vmCountLimit != 0 && totalExistingVMCount >= r.vmCountLimit {
+		r.log.Info("VM limit is already reached", "Limit", r.vmCountLimit, "Existing count", totalExistingVMCount)
+	}
+	r.vmCount = totalExistingVMCount
+
```
lukas016, line 191: i recommend integrate vm limit count into calculation of machine class quantity. It can help us improve scheduling.

lukas016 (changes requested):
a lot of stuff can be improved.

pkg/resources/manager/manager.go
so-sahu: Done

pkg/resources/manager/functions.go
so-sahu: Moved it to manager

pkg/resources/manager/functions.go
so-sahu: changed it to uint64

provider/cmd/app/app.go
so-sahu: type conversion is no longer needed here.

pkg/resources/manager/manager.go
```diff
@@ -437,6 +475,16 @@ func (r *resourceManager) getMachineClassAvailibilityAsString() string {
 	return removeSeparatorFromEnd(result)
 }
 
+func (r *resourceManager) getAvailableVMSlotsAsString() string {
+	count := r.availableVMSlots
+
+	if count < 0 {
```
lukas016, line 481: cannot be this little confusing? availablevmslots can be negative what i saw and this value is function is mainly use for debugging.

so-sahu: It can be negative if there are existing machines in store and the libvirt-provider is initialized with a lower VMLimit. However, in such cases, detailed information is available during the initialization of the resource-manager, which provides clarity to the backend team.

The `availableVMSlots` variable indicates the number of VMs that can still be created on the host. Therefore, if the limit is already reached, it should be set to 0.

lukas016 (approved):

PR #293, opened 2024-04-20
implement sgx

### LINKED ISSUE #157: Add `SGX` memory support
# Summary

This issue captures the efforts to implement `SGX` memory support and the corresponding reconciler changes. 
The proposal for the `api.Machine` spec changes will be discussed in this issue.

lukas016:
Fixes #157

### DIFF
--- a/pkg/resources/manager/functions.go
+++ b/pkg/resources/manager/functions.go
@@ -13,6 +13,7 @@ import (
 	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
 	"github.com/ironcore-dev/libvirt-provider/pkg/api"
 	"github.com/ironcore-dev/libvirt-provider/pkg/resources/sources"
+	"github.com/ironcore-dev/libvirt-provider/pkg/sgx"
 )
 
 // AddSource just registers source into manager
@@ -51,9 +52,14 @@ func SetLogger(logger logr.Logger) error {
 	return mng.setLogger(logger)
 }
 
-// SetMachineClasses just registers supported machineclasses
-func SetMachineClasses(classes []*iri.MachineClass) error {
-	return mng.setMachineClasses(classes)
+// SetMachineClasses just registers filename with machineclasses which are load during initializating
+func SetMachineClassesFilename(filename string) error {
+	return mng.setMachineClassesFilename(filename)
+}
+
+// GetIRIMAchineClasses will return machineClasses of resource manager as IRI Machine Classes
+func GetIRIMachineClasses() []iri.MachineClass {
+	return mng.getIRIMachineClasses()
 }
 
 // SetVMLimit just registers maximum limit for VMs
@@ -96,13 +102,15 @@ func GetSource(name string, options sources.Options) (Source, error) {
 		return sources.NewSourceCPU(options), nil
 	case sources.SourceHugepages:
 		return sources.NewSourceHugepages(options), nil
+	case sgx.SourceSGX:
+		return sgx.NewSourceSGX(options), nil
 	default:
 		return nil, fmt.Errorf("unsupported source %s", name)
 	}
 }
 
 func GetSourcesAvailable() []string {
-	return []string{sources.SourceCPU, sources.SourceMemory, sources.SourceHugepages}
+	return []string{sources.SourceCPU, sources.SourceMemory, sources.SourceHugepages, sgx.SourceSGX}
 }
 
 func GetMachineClassRequiredResources(name string) (core.ResourceList, error) {
@@ -111,7 +119,7 @@ func GetMachineClassRequiredResources(name string) (core.ResourceList, error) {
 		return nil, err
 	}
 
-	return class.resources.DeepCopy(), nil
+	return class.Capabilities.DeepCopy(), nil
 }
 
 func ValidateOptions(options sources.Options) error {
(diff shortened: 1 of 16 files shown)


pkg/resources/manager/manager.go
```diff
@@ -131,38 +133,60 @@ func (r *resourceManager) setVMLimit(maxVMsLimit uint64) error {
...
+		class.iriClass = &iri.MachineClass{
+			Name: class.Name,
+			Capabilities: &iri.MachineClassCapabilities{
+				CpuMillis:   cpu.MilliValue(),
+				MemoryBytes: mem.Value(),
 			},
 		}
 
+		for key := range class.Capabilities {
+			_, ok := r.registredResources[key]
```
so-sahu, line 164: One of the keys in capabilities is `memory_bytes`, which is not a resource.

pkg/resources/manager/manager.go
```diff
@@ -197,6 +221,10 @@ func (r *resourceManager) initialize(ctx context.Context, machines []*api.Machin
 			return err
 		}
 
+		for _, value := range resources.UnsortedList() {
+			r.registredResources[value] = s
```
so-sahu, line 225: If a resource is managed by multiple sources, it will be overwritten without warning. Ideally, an error should be thrown in such cases.

pkg/resources/manager/manager.go
```diff
@@ -257,18 +281,26 @@ func (r *resourceManager) allocate(machine *api.Machine, requiredResources core.
...
+		allocatedRes, err = s.Allocate(requiredResources)
+		if err != nil {
+			break
 		}
+
 		mergeResourceLists(totalAllocatedRes, allocatedRes)
 	}
 
+	if err != nil {
+		r.deallocateUnassignResources(totalAllocatedRes)
```
so-sahu, line 300: Can't it be moved to line 293, in place of break ?

pkg/resources/manager/manager.go
```diff
@@ -326,6 +361,12 @@ func (r *resourceManager) deallocate(machine *api.Machine, deallocateResources c
 	return nil
 }
 
+func (r *resourceManager) deallocateUnassignResources(resources core.ResourceList) {
+	for _, s := range r.sources {
```
so-sahu, line 365: Can't we use here the same mechanism that is being used above `s, ok := r.registredResources[key]` , in place of iterating over all the sources?

pkg/resources/manager/manager.go
```diff
@@ -360,18 +398,22 @@ func (r *resourceManager) getAvailableMachineClasses() []*iri.MachineClassStatus
...
 
-func (r *resourceManager) calculateMachineClassQuantity(class *machineclass) error {
+func (r *resourceManager) calculateMachineClassQuantity(class *MachineClass) error {
 	var count int64 = math.MaxInt64
 
-	for key, classQuantity := range class.resources {
-		available, err := getResource(key, r.getAvailableResources())
-		if err != nil {
-			return err
+	for _, s := range r.sources {
```
so-sahu, line 404: Can't we use here the same mechanism that is being used above s, ok := r.registredResources[key] , in place of iterating over all the sources?

pkg/resources/manager/manager.go
lukas016: hm? capabilities is resourcelist from machineclass file, it isn't iri.MachineClass

pkg/resources/manager/functions.go
```diff
@@ -111,7 +119,7 @@ func GetMachineClassRequiredResources(name string) (core.ResourceList, error) {
 		return nil, err
 	}
 
-	return class.resources.DeepCopy(), nil
+	return class.Capabilities.DeepCopy(), nil
```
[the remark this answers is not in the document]
lukas016: i use capabilities for keep compatibility with https://github.com/ironcore-dev/ironcore/blob/main/api/compute/v1alpha1/machineclass_types.go#L25

Technically for us it is still resourcelist with resources cannot it be more confusing? When i rename it, it can create confusing behavior because iri.MachineClass has capabilities too which are different.

pkg/resources/manager/manager.go
lukas016: we have this logic on line 233. This part of code will need to refactor.

pkg/resources/manager/manager.go
```diff
@@ -530,3 +572,21 @@ func removeSeparatorFromEnd(in string) string {
 
 	return in[:len(in)-len(sep)]
 }
+
+func loadMachineClasses(reader io.Reader) ([]*MachineClass, error) {
```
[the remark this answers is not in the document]
lukas016: i don't see big benefits in MCR anymore and i kept it for compatibility with upstream. I moved here because MachineClass is type under resource manager so i wanted keep it together.

pkg/resources/manager/functions.go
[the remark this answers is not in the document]
lukas016: discussed offline

pkg/resources/manager/manager.go
lukas016: fixed

pkg/resources/manager/manager.go
lukas016: fixed

pkg/resources/manager/manager.go
lukas016: fixed

so-sahu (approved):

PR #357, opened 2024-07-25
Implement ListEvents Runtime Interface to Support Cross-Cluster Events

### LINKED ISSUE #356: Implement `ListEvents` Runtime Interface to Support Cross-Cluster Events
# Summary

This proposal aims to implement the Runtime interface to support cross-cluster events, enhancing observability and debug-ability (see also https://github.com/ironcore-dev/ironcore/issues/1070).

## Basic example

Events of the `libvirt-provider` can be stored in-memory until they are fetched by the `poollet`. To avoid exceeding memory limits, events should be garbage collected after a defined TTL and the event store should evict the oldest events in case of reaching the max capacity.

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/356


cmd/libvirt-provider/app/app.go
```diff
@@ -148,6 +153,10 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.GCVMGracefulShutdownTimeout, "gc-vm-graceful-shutdown-timeout", 5*time.Minute, "Duration to wait for the VM to gracefully shut down. If the VM does not shut down within this period, it will be forcibly destroyed by garbage collector.")
 	fs.DurationVar(&o.ResyncIntervalGarbageCollector, "gc-resync-interval", 1*time.Minute, "Interval for resynchronizing the garbage collector.")
 
+	fs.IntVar(&o.MaximumMachineEvents, "maximum-machine-events", 100, "Maximum number of machine events that can be stored.")
```
lukas016, line 156: i recommend keep naming convention general -> specific. machine-event-max-events
Same i recommend do for names of vars.

internal/controllers/machine_controller.go
```diff
@@ -244,6 +252,9 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
 				}
 
 				if lastVolumeSize := getLastVolumeSize(machine, GetUniqueVolumeName(plugin.Name(), volumeID)); volumeSize != ptr.Deref(lastVolumeSize, 0) {
+					if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeNormal, "SizeChangedVolume", fmt.Sprintf("Volume size changed %s. lastVolumeSize: %d. volumeSize: %d .", volume.Name, *lastVolumeSize, volumeSize)); eventErr != nil {
```
lukas016, line 255: why . and you have different space for .

internal/controllers/machine_controller.go
```diff
@@ -530,11 +550,17 @@ func (r *MachineReconciler) updateDomain(
 
 	volumeStates, err := r.attachDetachVolumes(ctx, log, machine, attacher)
 	if err != nil {
+		if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeWarning, "AttchDetachVolume", fmt.Sprintf("Volume attach/detach failed with error: %s .", err)); eventErr != nil {
```
lukas016, line 553: do you need dot on end of line?

internal/controllers/machine_controller.go
```diff
@@ -698,15 +724,19 @@ func (r *MachineReconciler) domainFor(
...
 		}
 	}
 
 	if ignitionSpec := machine.Spec.Ignition; ignitionSpec != nil {
-		if err := r.setDomainIgnition(machine, domainDesc); err != nil {
+		if err := r.setDomainIgnition(log, machine, domainDesc); err != nil {
 			return nil, nil, nil, err
 		}
+	} else {
+		if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeWarning, "NoIgnitionData", "Machine does not have ignition data"); eventErr != nil {
```
lukas016, line 737: this event doesn't have dot on end of line. I personally prefer this format. How k8s doing that?

internal/controllers/machine_controller.go
```diff
@@ -716,13 +746,25 @@ func (r *MachineReconciler) domainFor(
...
+	}
 
 	nicStates, err := r.setDomainNetworkInterfaces(ctx, machine, domainDesc)
 	if err != nil {
+		if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeWarning, "AttchDetachNIC", fmt.Sprintf("Setting domain network interface failed with error: %s .", err)); eventErr != nil {
+			log.Error(eventErr, "failed to add machine event")
+		}
 		return nil, nil, nil, err
 	}
+	if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeNormal, "AttchedNIC", "Successfully attached network interfaces."); eventErr != nil {
```
lukas016, line 765: if machine doesn't have network interface, will this event generate too?

internal/server/event_list.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		iriEvents = append(iriEvents, iriEvent)
+	}
+	return iriEvents, nil
+}
+
+func (s *Server) ListEvents(ctx context.Context, req *iri.ListEventsRequest) (*iri.ListEventsResponse, error) {
+	iriEvents, err := s.listEvents()
+	if err != nil {
+		return nil, fmt.Errorf("error listing machine events : %w", err)
```
lukas016, line 62: delete space before :

internal/server/event_list.go
```diff
@@ -0,0 +1,70 @@
...
+		res = append(res, iriEvent)
+	}
+	return res
+}
+
+func (s *Server) listEvents() ([]*irievent.Event, error) {
+	events := s.eventStore.ListEvents()
+
+	var iriEvents []*irievent.Event
+	for _, event := range events {
```
lukas016, line 44: i think this require comment because ListEvents return []*irievent.Event type too

lukas016 (changes requested):

internal/server/event_list.go
lukas016: i don't see big benefits in this code. You break reference in evenStore.ListEvents(). Why do you have this for loop here?

lukas016: whole this function doesn't make sense. You can cool eventStore.ListEvents directly on line 60

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,124 @@
...
+func (es *EventStore) AddEvent(apiMetadata api.Metadata, eventType, reason, message string) error {
+	es.mutex.Lock()
+	defer es.mutex.Unlock()
+
+	// Calculate the index where the new event will be inserted
+	index := (es.head + es.count) % es.maxEvents
+
+	// If the store is full, log and overwrite the oldest event and move the head
+	if es.count == es.maxEvents {
+		es.log.V(1).Info("Overriding event", "event", es.events[es.head])
```
lukas016, line 51: is it debug level?

so-sahu: yes

internal/controllers/machine_controller.go
so-sahu: I referred to a Kubernetes document where the message ended with a period ("."). However, this seems to apply only to that specific object. I tested it for a pod, and the message doesn't end with a period. I will remove the period.

internal/controllers/machine_controller.go
so-sahu: I will remove the period. Thanks

internal/server/event_list.go
so-sahu: I agree with you. I had broken the reference in eventStore.ListEvents and forgot to remove it from here. I will take it out. Thanks for pointing that out 🙂.

internal/server/event_list.go
so-sahu: I will remove this code since it is no longer needed.

internal/server/event_list.go
so-sahu: removed

internal/server/event_list.go
so-sahu: removed

internal/controllers/machine_controller.go
so-sahu: Changed it so that it will only be printed when networkInterface is not nil. The same applies to volumes.

internal/controllers/machine_controller.go
so-sahu: Since Kubernetes follows the same convention, I adopted it as well.

internal/controllers/machine_controller.go
so-sahu: removed

internal/controllers/machine_controller.go
so-sahu: removed

cmd/libvirt-provider/app/app.go
so-sahu: applied.

internal/controllers/machine_controller.go
```diff
@@ -244,6 +252,9 @@ func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context
 				}
 
 				if lastVolumeSize := getLastVolumeSize(machine, GetUniqueVolumeName(plugin.Name(), volumeID)); volumeSize != ptr.Deref(lastVolumeSize, 0) {
+					if eventErr := r.eventStore.AddEvent(machine.Metadata, corev1.EventTypeNormal, "SizeChangedVolume", fmt.Sprintf("Volume size changed %s. lastVolumeSize: %d. volumeSize: %d", volume.Name, *lastVolumeSize, volumeSize)); eventErr != nil {
```
lukas016, line 255: could you pls maybe add here unit.

internal/controllers/machine_controller.go
lukas016: are you sure.

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,124 @@
...
+// ListEvents returns a copy of all events currently in the store.
+func (es *EventStore) ListEvents() []*irievent.Event {
+	es.mutex.Lock()
+	defer es.mutex.Unlock()
+
+	result := make([]*irievent.Event, es.count)
+	for i := 0; i < es.count; i++ {
+		index := (es.head + i) % es.maxEvents
+		// Create a deep copy of the event to break the reference
+		result[i] = proto.Clone(es.events[index]).(*irievent.Event)
```
lukas016, line 120: i am not big fun of retyping. Can it be implemented better?

internal/controllers/machine_controller.go
lukas016: and maybe it will be better replace . with comma because it looks as list of attributes

internal/event/machineevent/machineevent.go
so-sahu: done

internal/controllers/machine_controller.go
so-sahu: done

internal/controllers/machine_controller.go
so-sahu: done

internal/event/machineevent/machineevent_test.go
```diff
@@ -0,0 +1,148 @@
...
+	Context("RemoveExpiredEvents", func() {
+		It("should remove events whose TTL has expired", func() {
+			err := es.AddEvent(apiMetadata, eventType, reason, message)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(es.ListEvents()).To(HaveLen(1))
+
+			Eventually(func(g Gomega) bool {
+				es.RemoveExpiredEvents()
+				return g.Expect(es.ListEvents()).To(HaveLen(0))
+			}, eventTTL+1*time.Second, 100*time.Millisecond).Should(BeTrue())
```
lukas016, line 95: pls use Eventually with methods: WithTimeout(TIMEOUT).WithPolling(POLLING_INTERVAL).
It is better readable as just use time arguments without context.

lukas016: you have this problem on multiple places

internal/event/machineevent/machineevent_test.go
```diff
@@ -0,0 +1,148 @@
...
+		It("should return a copy of events", func() {
+			err := es.AddEvent(apiMetadata, eventType, reason, message)
+			Expect(err).ToNot(HaveOccurred())
+			events := es.ListEvents()
+			Expect(events).To(HaveLen(1))
+
+			events[0].Spec.Message = "Changed Message"
+
+			storedEvents := es.ListEvents()
+			Expect(storedEvents[0].Spec.Message).To(Equal(message))
```
lukas016, line 145: this was little worse readable for me. Maybe more straightforward will be
Expect(storeEvents[0].Spec.Message).ToNot(Equal(events[0].Spec.Message))

it is just cosmetic changes but i couldn't easy connect change message in events with this final comparison.

so-sahu: changed.

internal/event/machineevent/machineevent_test.go
so-sahu: Updated both instances of the `Eventually` block.

lukas016 (dismissed):
LGTM

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,129 @@
...
+
+		// Clear the reference to the expired event
+		es.events[index] = nil
+		es.head = (es.head + 1) % es.maxEvents
+		es.count--
+	}
+}
+
+// Start initializes and starts the event store's TTL expiration check.
+func (es *EventStore) Start(ctx context.Context, setupLog logr.Logger, machineEventResyncInterval time.Duration) {
```
lukasfrank, line 102: Is there a reason for leaking the option into the start method? 
Please use just `Start(ctx context.Context) error`?

lukasfrank (changes requested):

internal/controllers/machine_controller.go
```diff
@@ -99,6 +102,7 @@ func NewMachineReconciler(
 		libvirt:                        libvirt,
 		machines:                       machines,
 		machineEvents:                  machineEvents,
+		eventStore:                     eventStore,
```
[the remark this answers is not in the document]
so-sahu: Sure. I created the `EventRecorder` interface and replaced the direct usage of the event store in both the machine reconciler and the server struct with this interface. To accommodate this, I included both the `Eventf` and `ListEvents` methods in the `EventRecorder` interface.

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,129 @@
...
+			EventTime:          time.Now().Unix(),
+		},
+	}
+
+	es.events[index] = event
+	return nil
+}
+
+// RemoveExpiredEvents checks and removes events whose TTL has expired.
+func (es *EventStore) RemoveExpiredEvents() {
```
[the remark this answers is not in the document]
so-sahu: Done

internal/event/machineevent/machineevent_test.go
```diff
@@ -0,0 +1,148 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(es.ListEvents()).To(HaveLen(i + 1))
+			}
+
+			err := es.AddEvent(apiMetadata, eventType, reason, "New Event")
+			Expect(err).ToNot(HaveOccurred())
+
+			events := es.ListEvents()
+			Expect(events).To(HaveLen(maxEvents))
+			Expect(events[maxEvents-1].Spec.Message).To(Equal("New Event"))
```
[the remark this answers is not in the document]
so-sahu: Done

cmd/libvirt-provider/app/app.go
```diff
@@ -305,7 +314,7 @@ func Run(ctx context.Context, opts Options) error {
 		return err
 	}
 
-	machineEvents, err := event.NewListWatchSource[*api.Machine](
```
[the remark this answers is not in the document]
so-sahu: Sure. Now that we're referring to the machine event store as `EventRecorder`, it eliminates any confusion with machine events themselves. Therefore, I've renamed it back to `machineEvents`.

internal/event/machineevent/machineevent.go
so-sahu: Initially, I intended to log the shutdown of the event garbage collector using the setup log. However, since none of the `Start` methods include this, I've removed it. I've also simplified the definition to `Start(ctx context.Context)` since there's no error handling required in this case.

internal/controllers/machine_controller.go
```diff
@@ -131,6 +135,7 @@ type MachineReconciler struct {
 
 	machines      store.Store[*api.Machine]
 	machineEvents event.Source[*api.Machine]
+	eventRecorder machineEvent.EventRecorder
```
lukasfrank, line 138: Maybe just embed the EventRecorder in order to let us call it directly like `r.Eventf(....) `
```suggestion
	machineEvent.EventRecorder
```

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,140 @@
...
+		eventTTL:            opts.MachineEventTTL,
+		eventResyncInterval: opts.MachineEventResyncInterval,
+		head:                0,
+		count:               0,
+		log:                 log,
+	}
+}
+
+// Eventf adds a new Event to the store. Implements the EventRecorder interface.
+func (es *EventStore) Eventf(apiMetadata api.Metadata, eventType, reason, message string) error {
```
lukasfrank, line 57: Can we put the metadata directly in the recording function that we do not need to return an error?

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/gogo/protobuf/proto"
+	irievent "github.com/ironcore-dev/ironcore/iri/apis/event/v1alpha1"
+	"github.com/ironcore-dev/libvirt-provider/api"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+type EventRecorder interface {
+	Eventf(apiMetadata api.Metadata, eventType, reason, message string) error
```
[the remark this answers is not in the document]
lukasfrank: Can we please stick to the naming what is used in controller runtime? 
https://github.com/ironcore-dev/libvirt-provider/pull/357#discussion_r1701940919

internal/controllers/machine_controller.go
so-sahu: Done

internal/event/machineevent/machineevent.go
so-sahu: Removed the return type for errors, as event recording should not disrupt the main business logic. Instead, errors are now logged directly without being returned.

internal/event/machineevent/machineevent.go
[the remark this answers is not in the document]
so-sahu: Done. Now the Eventf supports string fmt.

internal/server/event_list.go
lukasfrank, line 1: It would be great to have here one or two tests to check if events can be fetched/filtered

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/ironcore-dev/libvirt-provider/api"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// Interface for recording events
+type EventRecorder interface {
+	Eventf(log logr.Logger, apiMetadata api.Metadata, eventType string, reason string, messageFormat string, args ...any)
+}
+
+// Interface for listing events
```
lukasfrank, line 25: Comment should have the following format 'EventStore ...' (with an optional leading article)

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,154 @@
...
+	mutex               sync.Mutex        // Mutex for thread safety
+	eventTTL            time.Duration     // TTL for events
+	eventResyncInterval time.Duration     // Resync interval for event store's TTL expiration check
+	head                int               // Index of the oldest event
+	count               int               // Current number of events in the store
+	log                 logr.Logger       // Logger for logging overridden events
+}
+
+// NewEventStore creates a new EventStore with a fixed number of events and set TTL for events.
+func NewEventStore(log logr.Logger, opts EventStoreOptions) *eventStore {
```
lukasfrank, line 50: Exported function with the unexported return type. Maybe change `eventStore` to `Store`. 
I think together with the package import it's still clear what we are using (`machineevent.Store`)

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,154 @@
...
+	Eventf(log logr.Logger, apiMetadata api.Metadata, eventType string, reason string, messageFormat string, args ...any)
+}
+
+// Interface for listing events
+type EventStore interface {
+	ListEvents() []*irievent.Event
+}
+
+// Options to initialize the machine event store
+type EventStoreOptions struct {
```
lukasfrank, line 31: Comment should have the following format 'EventStoreOptions ...' (with an optional leading article)

internal/event/machineevent/machineevent.go
```diff
@@ -0,0 +1,154 @@
...
+	MachineEventMaxEvents      int
+	MachineEventTTL            time.Duration
+	MachineEventResyncInterval time.Duration
+}
+
+// eventStore implements the EventRecorder interface and represents an in-memory event store with TTL for events.
```
lukasfrank, line 37: eventStore implements the EventRecorder and EventStore interface and represents an in-memory event store with TTL for events.

lukasfrank (approved):
Thanks a lot for the PR, great work!

PR #370, opened 2024-08-13
Extend Server Tests to Cover ListEvents

### LINKED ISSUE #366: Extend the `Server` tests to cover `ListEvents`
It would be great to have here one or two tests to check if events can be fetched/filtered

_Originally posted by @lukasfrank in https://github.com/ironcore-dev/libvirt-provider/pull/357#discussion_r1706478258_

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/366

### DIFF
--- a/internal/server/event_list_test.go
+++ b/internal/server/event_list_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/digitalocean/go-libvirt"
+	irievent "github.com/ironcore-dev/ironcore/iri/apis/event/v1alpha1"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/machine/v1alpha1"
 	irimeta "github.com/ironcore-dev/ironcore/iri/apis/meta/v1alpha1"
 	machinepoolletv1alpha1 "github.com/ironcore-dev/ironcore/poollet/machinepoollet/api/v1alpha1"
@@ -116,5 +117,12 @@ var _ = Describe("ListEvents", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		Expect(resp.Events).To(BeEmpty())
+
+		By("listing the machine events with expired TTL")
+		Eventually(func(g Gomega) []*irievent.Event {
+			resp, err := machineClient.ListEvents(ctx, &iri.ListEventsRequest{})
+			g.Expect(err).NotTo(HaveOccurred())
+			return resp.Events
+		}).Should(BeEmpty())
 	})
 })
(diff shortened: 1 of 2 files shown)


internal/server/event_list_test.go
```diff
@@ -0,0 +1,126 @@
...
+		domainXMLData, err := libvirtConn.DomainGetXMLDesc(domain, 0)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(domainXMLData).NotTo(BeEmpty())
+
+		By("ensuring domain for machine is in running state")
+		Eventually(func(g Gomega) libvirt.DomainState {
+			domainState, _, err := libvirtConn.DomainGetState(domain, 0)
+			g.Expect(err).NotTo(HaveOccurred())
+			return libvirt.DomainState(domainState)
+		}).Should(Equal(libvirt.DomainRunning))
```
lukasfrank, line 66: Don't we repeat our selfs here? This block seems to be present in every test. I propose to drop it since we test this functionality somewhere else.

internal/server/event_list_test.go
```diff
@@ -0,0 +1,126 @@
...
+			Eventually(func(g Gomega) bool {
+				_, err := machineClient.DeleteMachine(ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})
+				g.Expect(err).To(SatisfyAny(
+					BeNil(),
+					MatchError(ContainSubstring("NotFound")),
+				))
+				_, err = libvirtConn.DomainLookupByUUID(libvirtutils.UUIDStringToBytes(createResp.Machine.Metadata.Id))
+				return libvirt.IsNotFound(err)
+			}).Should(BeTrue())
+		})
```
lukasfrank, line 49: Please do not wrap this in a Eventually block: 

```suggestion
		DeferCleanup(machineClient.DeleteMachine, ctx, &iri.DeleteMachineRequest{MachineId: createResp.Machine.Metadata.Id})		
```

lukasfrank (changes requested):

internal/server/event_list_test.go
so-sahu: Removed

internal/server/event_list_test.go
so-sahu: Done

so-sahu: As discussed, we are sticking with the current implementation due to the dependency issue. I have created a follow-up issue here: https://github.com/ironcore-dev/libvirt-provider/issues/380.

lukasfrank (approved):

PR #384, opened 2024-08-26
Implemented reconcile metrics

KirrTap:
Added metrics to the reconcilers

### DIFF
--- a/internal/controllers/machine_controller.go
+++ b/internal/controllers/machine_controller.go
@@ -15,8 +15,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/digitalocean/go-libvirt"
+	libvirt "github.com/digitalocean/go-libvirt"
 	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
+	"libvirt.org/go/libvirtxml"
+
 	"github.com/ironcore-dev/libvirt-provider/api"
 	"github.com/ironcore-dev/libvirt-provider/internal/event"
 	machineEvent "github.com/ironcore-dev/libvirt-provider/internal/event/machineevent"
@@ -32,12 +35,11 @@ import (
 	"github.com/ironcore-dev/libvirt-provider/internal/raw"
 	"github.com/ironcore-dev/libvirt-provider/internal/store"
 	"github.com/ironcore-dev/libvirt-provider/internal/utils"
-	"github.com/prometheus/client_golang/prometheus"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/ptr"
-	"libvirt.org/go/libvirtxml"
 )
 
 const (
@@ -52,20 +54,18 @@ const (
 	MachineReconcilerOpsLibvirtEvent     = "libvirt-event"
 )
 
-var (
-	// TODO: improve domainStateToMachineState since some states are mapped to computev1alpha1.MachineStatePending
-	// where it doesn't make that much sense.
-	domainStateToMachineState = map[libvirt.DomainState]api.MachineState{
-		libvirt.DomainNostate:  api.MachineStatePending,
-		libvirt.DomainRunning:  api.MachineStateRunning,
-		libvirt.DomainBlocked:  api.MachineStatePending,
-		libvirt.DomainPaused:   api.MachineStatePending,
-		libvirt.DomainShutdown: api.MachineStateTerminating,
-		// it isn't probably supported by transient domain
-		libvirt.DomainShutoff:     api.MachineStateTerminated,
-		libvirt.DomainPmsuspended: api.MachineStatePending,
-	}
-)
+// TODO: improve domainStateToMachineState since some states are mapped to computev1alpha1.MachineStatePending
+// where it doesn't make that much sense.
+var domainStateToMachineState = map[libvirt.DomainState]api.MachineState{
+	libvirt.DomainNostate:  api.MachineStatePending,
+	libvirt.DomainRunning:  api.MachineStateRunning,
+	libvirt.DomainBlocked:  api.MachineStatePending,
+	libvirt.DomainPaused:   api.MachineStatePending,
+	libvirt.DomainShutdown: api.MachineStateTerminating,
+	// it isn't probably supported by transient domain
+	libvirt.DomainShutoff:     api.MachineStateTerminated,
+	libvirt.DomainPmsuspended: api.MachineStatePending,
+}
 
 type MachineReconcilerOptions struct {
 	GuestCapabilities                       guest.Capabilities
@@ -169,7 +169,7 @@ type MachineReconciler struct {
 func (r *MachineReconciler) Start(ctx context.Context) error {
 	log := r.log
 
-	//todo make configurable
+	// todo make configurable
 	workerSize := 15
 	metrics.ControllerRuntimeMaxConccurrentReconciles.WithLabelValues(MachineReconcilerName).Set(float64(workerSize))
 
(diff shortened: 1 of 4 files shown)


internal/controllers/machine_controller.go
```diff
@@ -221,9 +235,20 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 }
 
 func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context, log logr.Logger) {
+	const name = "volume-size"
+	log = log.WithName(name)
+	opsDuration := metrics.OperationDuration.WithLabelValues(name)
+	opsErrors := metrics.OperationErrors.WithLabelValues(name)
```
lukasfrank, line 241: Can we keep passing the prepared logger and maybe define a global name like you did for `MachineReconcilerName`?

internal/controllers/machine_controller.go
```diff
@@ -145,13 +154,18 @@ type MachineReconciler struct {
...
+	metricsControllerRuntimeActiveWorker    prometheus.Gauge
+	metricsControllerRuntimeReconcileErrors prometheus.Counter
 }
 
 func (r *MachineReconciler) Start(ctx context.Context) error {
 	log := r.log
 
 	//todo make configurable
 	workerSize := 15
+	metrics.ControllerRuntimeMaxConccurrentReconciles.WithLabelValues("machine-reconciler").Set(float64(workerSize))
```
lukasfrank, line 168: pls use the const here

internal/controllers/machine_controller.go
```diff
@@ -98,24 +101,30 @@ func NewMachineReconciler(
...
+		volumePluginManager:                     opts.VolumePluginManager,
+		networkInterfacePlugin:                  opts.NetworkInterfacePlugin,
+		resyncIntervalVolumeSize:                opts.ResyncIntervalVolumeSize,
+		resyncIntervalGarbageCollector:          opts.ResyncIntervalGarbageCollector,
+		enableHugepages:                         opts.EnableHugepages,
+		gcVMGracefulShutdownTimeout:             opts.GCVMGracefulShutdownTimeout,
+		volumeCachePolicy:                       opts.VolumeCachePolicy,
+		metricsReconcileDuration:                metrics.ControllerRuntimeReconcileDuration.WithLabelValues(MachineReconcilerName),
+		metricsControllerRuntimeActiveWorker:    metrics.ControllerRuntimeActiveWorker.WithLabelValues(MachineReconcilerName),
+		metricsControllerRuntimeReconcileErrors: metrics.ControllerRuntimeReconcileErrors.WithLabelValues(MachineReconcilerName),
```
lukasfrank, line 127: Why are the metrics not passed as options? It makes it hard to pass a mock metric provider in tests

internal/controllers/machine_controller.go
```diff
@@ -98,24 +101,30 @@ func NewMachineReconciler(
 	}
 
 	return &MachineReconciler{
```
lukasfrank, line 103: Can we add at least a simple test?

lukasfrank (changes requested):

internal/controllers/machine_controller.go
lukas016: why do you want to mock metrics? these metrics has to represent internal stuff of controller (errors, workers, ...) and they don't have any special logic so you can create them in unit tests and it will work.

lukasfrank: You can pass your own simple implementation in order to check the functionality.

internal/controllers/machine_controller.go
lukas016: about passing argument i am not completely sure.  I personally suggest remove passing over argument and directly initialize in function log from r.log. because passing of arguments doesn't add any benefits because start.. functions are methods of MachineReconciler and they have direct access to main logger. Now we just "spilitting" configuration of operation into 2 functions. I prefer keep it together.

My suggestion:

```
func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context) {
  log := r.log.WithName(MachineReconcilerVolumeSize)
  ...
```

internal/controllers/machine_controller.go
lukas016: without support of mocking we can just create integration test.

internal/controllers/machine_controller.go
lukas016: i checked that and i don't think we need mocking interface.

You can very nice test metrics directly with testutil package from prometheus without web server:

```
var _ = Describe("Controller runtime metrics", func() {
	Context("metric ControllerRuntimeReconcileErrors", func() {
		It("should increment metric", func() {
			m := metrics.ControllerRuntimeReconcileErrors.WithLabelValues(controllers.MachineReconcilerName)
			m.Inc()
			m.Inc()
			expectedMetric := `
# HELP controller_runtime_reconcile_errors_total Total number of reconciliation errors per controller
# TYPE controller_runtime_reconcile_errors_total counter
controller_runtime_reconcile_errors_total{controller="machine-reconciler"} 2
`
			reader := strings.NewReader(expectedMetric)
			Expect(testutil.GatherAndCompare(prometheus.DefaultGatherer, reader, "controller_runtime_reconcile_errors_total")).ToNot(HaveOccurred())
		})
	})
})
```

internal/controllers/machine_controller.go
lukas016: we can implement both see my comment: https://github.com/ironcore-dev/libvirt-provider/pull/384#discussion_r1777179923

internal/controllers/machine_controller.go
lukasfrank: The code above describes a test case which checks if you can increment a metrics counter, which is covered already by the lib. 
I'm talking about a test where we create a resource, and check if metric `x`, or `y` is as expected.

internal/controllers/machine_controller.go
lukasfrank: `WithName` adds context and appends the name to the existing name.

lukas016: I really don't understand why are you giving me here function description and you don't react usefully to my comment.
But okay it appends name separate with `.`, if you want to see whole description.
But it is useless, if we discuss how passing child logger into function.

internal/controllers/machine_controller.go
lukas016: you have it described above.

internal/controllers/machine_controller.go
balpert89: Dissecting the requests here.

1. Request to introduce global consts for the individual logger decorators: I agree with @lukasfrank here, being consistent here is introducing a const for each decorator name similar to https://github.com/ironcore-dev/libvirt-provider/pull/384/files#diff-a2d4663c7af1f98625f6b53f602e53ce910de934bd827f232e31db91bf9bb5acR49

2. Injecting vs. Self-decorating: This is a bit of a matter of taste in this particular spot. You can argue for both sides, keeping the injection by method argument enables the method to just use whatever logger is passed in and go with it. Also, configuration is kept outside of the method itself and is done by the calling instance. The self-decorating approach @lukas016 suggests is also valid as it encapsulates the responsibilities for behavior in the `start...` methods. Ultimately, the suggestion by @lukas016 is a good compromise:

```
func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context) {
  log := r.log.WithName(MachineReconcilerVolumeSize)
  ...
```

When testing, you pass in your mock logging provider to the containing reconciler struct and the `start...` methods consume this, with also self-contained decorators.

internal/controllers/machine_controller.go
balpert89: applied in https://github.com/ironcore-dev/libvirt-provider/pull/384/commits/faaa29bd13c03b74f24feeedd931a2b98430a23b

lukas016:
i am closing issue because owners of repo doesn't collaborate with us.

PR #397, opened 2024-09-11
Fixed in-cluster `apinetClient`

nitrocb:
# Proposed Changes
- added in-cluster case for apinetCfg, as this was not covered before leading to a panic (nilpointer-deref) for in-cluster cases

### DIFF
--- a/cmd/libvirt-provider/app/app.go
+++ b/cmd/libvirt-provider/app/app.go
@@ -222,7 +222,6 @@ func Run(ctx context.Context, opts Options) error {
 	}
 
 	// Check if apinetKubeconfig is provided
-	var apinetClient client.Client
 	var apinetCfg *rest.Config
 	if opts.ApinetKubeconfig != "" {
 		apinetCfg, err = clientcmd.BuildConfigFromFlags("", opts.ApinetKubeconfig)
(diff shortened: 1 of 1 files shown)


cmd/libvirt-provider/app/app.go
```diff
@@ -222,20 +223,28 @@ func Run(ctx context.Context, opts Options) error {
 
 	// Check if apinetKubeconfig is provided
 	var apinetClient client.Client
```
lukasfrank, line 225: Do we still need this variable here?

cmd/libvirt-provider/app/app.go
```diff
@@ -222,20 +223,28 @@ func Run(ctx context.Context, opts Options) error {
...
+		// assuming in-cluster config
+		apinetCfg, err = rest.InClusterConfig()
 		if err != nil {
-			setupLog.Error(err, "failed to initialize api-net client")
+			setupLog.Error(err, "failed to create in-cluster-config")
 			return err
 		}
 	}
 
+	apinetClient, err = client.New(apinetCfg, client.Options{Scheme: scheme})
```
lukasfrank, line 242:
```suggestion
	apinetClient, err := client.New(apinetCfg, client.Options{Scheme: scheme})
```

cmd/libvirt-provider/app/app.go
nitrocb: As the client creation is not part of the if condition anymore, we don't need that variable anymore, too.
Thanks for that hint! I have removed it.

lukasfrank (approved):
Thanks a lot for your contribution! LGTM

PR #406, opened 2024-09-23
Added Configurable Volume Caching Policies

### LINKED ISSUE #389: Add configurable volume caching policies
# Summary

Currently our volume drivers are configured without any caching enabled. We should allow to influence how volume caching is being used for remote attached disks.

In domain XML terms this means we need to influence the `driver` `cache` field :

```
<disk type='network' device='disk'>
  <driver name='qemu' type='raw' cache='writeback'/>
  <source protocol='rbd' name='pool_name/volume_name'>
    <host name='ceph-mon1' port='6789'/>
    <host name='ceph-mon2' port='6789'/>
    <host name='ceph-mon3' port='6789'/>
  </source>
  <target dev='vda' bus='virtio'/>
  <address type='pci' domain='0x0000' bus='0x00' slot='0x04' function='0x0'/>
</disk>
```

We could expose this configuration via a new flag e.g. `volumeCachePolicy` which indicates which policy to use when creating a disk. The default should be `none`.

so-sahu:
Fixes https://github.com/ironcore-dev/libvirt-provider/issues/389

### DIFF
--- a/cmd/libvirt-provider/app/app.go
+++ b/cmd/libvirt-provider/app/app.go
@@ -93,6 +93,8 @@ type Options struct {
 	ResyncIntervalGarbageCollector time.Duration
 
 	MachineEventStore machineevent.EventStoreOptions
+
+	VolumeCachePolicy string
 }
 
 type HTTPServerOptions struct {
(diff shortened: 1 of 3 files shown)


cmd/libvirt-provider/app/app.go
```diff
@@ -156,6 +158,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.MachineEventStore.MachineEventTTL, "machine-event-ttl", 5*time.Minute, "Time to live for machine events.")
 	fs.DurationVar(&o.MachineEventStore.MachineEventResyncInterval, "machine-event-resync-interval", 1*time.Minute, "Interval for resynchronizing the machine events.")
 
+	fs.StringVar(&o.VolumeCachePolicy, "volume-cache-policy", "none", "Policy to use when creating a remote disk. Default is none")
```
lukas016, line 161: what are other options?

so-sahu: Based on my understanding, the accepted values are `none`, `writeback`, and `writethrough`. I assumed that users setting this option explicitly would already be familiar with these values, so I didn’t add any additional validations or detailed information. Additionally, if more values are supported in the future, we would need to update the validation logic, which could introduce unnecessary maintenance overhead.

lukas016: validation isn't such important. but pls add minimally possible option what can be problematic because it depends on hypervisor and libvirtd version. Maybe you can add this note and use official docs: https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms here you have to scroll on attribute drive where are all options.

so-sahu: Added

lukas016 (approved):
LGTM

afritzler (approved):

lukasfrank (approved):

PR #407, opened 2024-09-23
fix

lukasfrank:
# Proposed Changes

-
-
-

Fixes #


lukas016:
it is breaking using of isolated and providernet network plugin too. Pls fix it too. This argument isn't required for other network plugins just for apinet.

PR #444, opened 2024-11-26
Fix RBD connection handling in `Volume` resize

lukas016:
# Proposed Changes

Close rdb connection after getting volume size.

### DIFF
--- a/internal/controllers/machine_controller.go
+++ b/internal/controllers/machine_controller.go
@@ -222,11 +222,12 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 
 func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context, log logr.Logger) {
 	if r.resyncIntervalVolumeSize == 0 {
+		log.V(1).Info("volume resize trigger loop is disabled")
 		return
 	}
 
 	wait.UntilWithContext(ctx, func(ctx context.Context) {
-		log.V(1).Info("starting volume resize loop")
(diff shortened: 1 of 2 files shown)


internal/controllers/machine_controller.go
```diff
@@ -221,7 +221,12 @@ func (r *MachineReconciler) Start(ctx context.Context) error {
 }
 
 func (r *MachineReconciler) startCheckAndEnqueueVolumeResize(ctx context.Context, log logr.Logger) {
+	if r.resyncIntervalVolumeSize == 0 {
+		return
```
lukasfrank, line 225: Should we print an info if the volume resize is disabled?

lukas016: sure we can.

lukas016: maybe not, this technically doesn't disable resize logic. It will just disable triggering of resize base on volume size. If this volume size check loop is deactivated, it doesn't mean resize is disabled. Resize will still executed in reconcile loop. Currently i think all machines are reconcile every hour, if you don't have any other trigger mechanism of reconcile loop.
So just simple message about volume resize is disabled will be more confusing for users because it isn't true.
Technically we can add logs: Reconciliation of volume size base on change of capacity (or size) is disabled.

But this won't be very useful for users too without context and we don't have any documentation where we describe how libvirt-provider is working.

lukas016: updated @lukasfrank

lukasfrank (approved):

afritzler (approved):

PR #506, opened 2025-02-20
Add support for `arm64` container build

### LINKED ISSUE #40: `linux/arm64` support
**Describe the bug**
Investigate how we can support `linux/arm64`

afritzler:
# Proposed Changes

Support the container image build on `arm64` architectures.

Fixes #40

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,20 +7,8 @@ ENVTEST_K8S_VERSION = 1.28
 # Docker image name for the mkdocs based local development setup
 MKDOCS_IMG=onmetal/libvirt-provider-docs
 
-# Code depend on OS
-TARGET_OS ?= linux
-TARGET_ARCH ?= amd64
-CGO_ENABLED ?= 1
-
 LIBVIRT_PROVIDER_BIN=$(LOCALBIN)/libvirt-provider
 LIBVIRT_PROVIDER_BIN_SOURCE=./cmd/libvirt-provider
-
-GITHUB_PAT_PATH ?=
-ifeq (,$(GITHUB_PAT_PATH))
-GITHUB_PAT_MOUNT ?=
-else
-GITHUB_PAT_MOUNT ?= --secret id=github_pat,src=$(GITHUB_PAT_PATH)
-endif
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -122,7 +108,7 @@ clean-docs: ## Remove all local mkdocs Docker images (cleanup).
 
 .PHONY: build
 build: manifests generate fmt vet add-license lint ## Build the binary
-	GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) CGO_ENABLED=$(CGO_ENABLED) go build -o $(LIBVIRT_PROVIDER_BIN) $(LIBVIRT_PROVIDER_BIN_SOURCE)
```
lukas016, line 125: why are you removing flexibility of makefile?

afritzler: To make the `Makefile` cleaner. `CGO` has to be always enabled. ARCH and OS are enforced by you local environment anyways and a local cross build without Docker won't be feasible due to the `lib` binding dependencies. So in short: keep the `make build` short and precise as this is only used for local dev builds.

lukas016: It isn't true in long term. when we delete dependency on rbd (ceph resize logic), you can use build target for cross platform build but i think global variables can still work. But CGO_ENABLED is defined on multiple places in makefile and it is better manage this value from one place as change it in future on multiple places.

afritzler: When we remove the `ceph` dependency the `CGO_ENGABLED` flag also needs to go as this is not needed.

Makefile
```diff
@@ -133,7 +119,7 @@ run: manifests generate fmt vet ## Run the binary
 
 .PHONY: docker-build
 docker-build: ## Build docker image with partitionlet
-	$(CONTAINER_TOOL) build $(CONTAINER_BUILDARGS) -t ${IMG} $(GITHUB_PAT_MOUNT) .
```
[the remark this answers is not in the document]
afritzler: Ok, I put the `CONTAINER_BUILDARGS` back in but determine the architecture dynamically based on the OS your environment. You can always overwrite it from the outside.

lukasfrank (approved):

PR #655, opened 2025-10-31
Added `LocalDisk`

### LINKED ISSUE #657: Introduction of `LocalDisk` for `Machine`s
# Summary

This feature should introduce a ephemeral`LocalDisk` for `Machine`s which is provided by the `MachinePool` provider.  It should  deprecate the `EmptyDiskVolumeSource` and the `image` in the `Machine`  

The `LocalDiskVolumeSource` should also provide the option to provide an `image` to pre-populate the `LocalDisk` with an operation system: 

```
type LocalDiskVolumeSource struct {
	SizeLimit *resource.Quantity 
	Image string
}
```

## Motivation

This change solves the following issue: The API allows to define the Device names which should be respected by the `libvirt-provider`. It makes it transparent which image is used to boot the machine.

lukasfrank:
# Proposed Changes

- Add `LocalDisk`
- Updated tests 

Fixes #657


gonzolino (approved):
lgtm

In the future, if we want to support other types of local storage (e.g. LVM disks), we might need to rename this.

Issue #678, opened 2026-01-19
Implement GPU passthrough support


lukasfrank:
# Summary

Enhance the `libvirt-provider` to support GPU passthrough for virtual machines. The provider should use https://github.com/ironcore-dev/provider-utils/issues/42 to claim the resources and use them respectively. 

- Extend the server to claim resources 
- Extend the API to support passthrough devices 
- Extend the controller 
  - to use the new API field(s) and use them in the domain xml 
  - andRelease the resources

PR #680, opened 2026-01-22
Implement Nvidia GPU passthrough

friegger:
- Claim GPUs during IRI machine create
- Controller adds GPU PCI addresses to domain XML for passthrough
- Controller releases claims on machine deletion
- Release claims in case of errors

This does not enable to use NVLink with mutiple GPUs.

Contributes to #678 

## Summary by CodeRabbit

* **New Features**
  - GPU allocation and management: machines can claim, expose, and use Nvidia GPUs; a resource claimer starts with the server and coordinates claims/releases. Server now requires a resource claimer on startup.

* **Configuration**
  - Machine class schema supports GPU resources alongside CPU and memory.
  - Machine specs persist GPU PCI addresses when GPUs are assigned.

* **Behavior**
  - Creating a machine will fail if requested GPUs are exhausted; claims are released on deletion.

* **Tests**
  - Added tests for GPU provisioning, claim exhaustion, PCI-address handling, and host-device conversion.


afritzler:
Just a small nitpick: Can we stick to PR titles which do not include the categorization e.g. feature etc? I would suggest to keep the release log as they are generated at the moment consistent: https://github.com/ironcore-dev/libvirt-provider/releases

lukasfrank (approved):

PR #681, opened 2026-01-28
Add hostname propagation to ironcore-net ` NetworkInterfaces`

### LINKED ISSUE #538: Add `hostname` propagation to `ironcore-net` `NetworkInterfaces`
# Summary

Annouce the `Machine` name as the hostname of the VM. This can be now achieved by propagating the hostname to the `ironcore-net` `NetworkInterface` via the `apinet` network plugin.

## Tasks
- Add hostname field when creating the `ironcore-net` `NetworkInterface`
- Ensure the correct behaviour by extending the existing test suite

/ref https://github.com/ironcore-dev/ironcore-net/pull/400

ushabelgur:
# Proposed Changes

- Pass machineName as hostaname to  ironcore-net ` NetworkInterfaces` while creating machine
- Update testcase
- Update ironcore-net dependency 

Fixes #538 

## Summary by CodeRabbit

* **New Features**
  * Machines can include an optional guest configuration with a hostname; network interfaces now receive and use hostnames from machine guest config to improve identification.

* **Chores**
  * Bumped Go toolchain and dependency versions and updated pinned tooling versions used for code generation and tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ironcore-dev/libvirt-provider/pull/681?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lukasfrank (changes requested):

api/machine.go
```diff
@@ -107,6 +107,7 @@ type NetworkInterfaceSpec struct {
 	NetworkId  string            `json:"networkId"`
 	Ips        []string          `json:"ips"`
 	Attributes map[string]string `json:"attributes"`
+	HostName   string            `json:"hostName,omitempty"`
```
ushabelgur: @lukasfrank @afritzler should I introduce `GuestConfig` type for libvirt machine type? or only ` HostName` is fine for now ?

friegger: I think it is a good idea to keep it consistent with the IRI side and introduce the GuestConfig. Though, not on the `NetworkInterfaceSpec`, but on the `MachineSpec`, like on the IRI side. Reasoning is, that there can be multiple NICs and with that we would introduce a possibility of a bug with inconsistent hostnames accross multiple NICs. Something that should be one, would be many. The apinet plugin has also access to the machine, so it can take the hostname there from the machine's GuestConfig. @lukasfrank What do you think?

ushabelgur: Changes incorporated to have `guestConfig` in `Machine`

internal/plugins/networkinterface/apinet/apinet.go
```diff
@@ -167,34 +168,33 @@ func (p *Plugin) Apply(ctx context.Context, spec *api.NetworkInterfaceSpec, mach
...
-		},
-		Spec: apinetv1alpha1.NetworkInterfaceSpec{
-			NetworkRef: corev1.LocalObjectReference{
-				Name: apinetNetworkName,
-			},
-			NodeRef: corev1.LocalObjectReference{
-				Name: p.nodeName,
-			},
-			IPs: ironcoreIPsToAPInetIPs(spec.Ips),
-		},
```
lukasfrank, line 190: Can't we keep the change small and add the single line to support the `HostName`? Is there a reason to drop the struct-literals?

ushabelgur: This change is needed because k8s 1.35 update, instead of Patch we are using Apply method now

lukasfrank (approved):

PR #740, opened 2026-07-05
Use native filtering of `store.Store`

gonzolino:
# Proposed Changes

- Use `MatchingLabels` an `MatchingFields` filters for listing instead of post-fetch loops

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved machine lookups during image handling and garbage collection by filtering results at the store level.
  * Limited machine listings to provider-managed machines for more efficient responses.

* **Bug Fixes**
  * Improved filtering of deleted machines and machines associated with specific images.

* **Tests**
  * Updated test coverage and setup to support the enhanced machine filtering behavior.

### DIFF
--- a/cmd/libvirt-provider/app/app.go
+++ b/cmd/libvirt-provider/app/app.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	claim "github.com/ironcore-dev/provider-utils/claimutils/claim"
+	"github.com/ironcore-dev/provider-utils/storeutils/store"
 
 	"github.com/go-logr/logr"
 	"github.com/ironcore-dev/ironcore-image/oci/remote"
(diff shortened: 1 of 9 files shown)


cmd/libvirt-provider/app/app.go
```diff
@@ -296,6 +297,11 @@ func Run(ctx context.Context, opts Options) error {
 		NewFunc:        func() *api.Machine { return &api.Machine{} },
 		CreateStrategy: strategy.MachineStrategy,
 		Dir:            providerHost.MachineStoreDir(),
+		FieldIndexers: map[string]store.IndexerFunc[*api.Machine]{
+			api.MachineMetadataDeletedField: api.SetupMachineMetadataDeletedFieldIndexer,
+			api.MachineSpecImageField:       api.SetupMachineSpecImageFieldIndexer,
+			api.MachineSpecHasGpuField:      api.SetupMachineSpecHasGpuFieldIndexer,
```
gonzolino: Note: I am unsure if it is worthwhile to add an index for `api.MachineSpecHasGpuField`, as this index is only used once during startup in `GetClaimedPCIAddressesFromMachineStore`.

lukasfrank: I have no strong opinion on that but would lean slightly into the direction and drop it

lukasfrank (changes requested):
please update also the tests, that we have no production/test divergence.  (use the same indexers there)

cmd/libvirt-provider/app/app.go
gonzolino: Ok, since we both tend to drop it, I removed the index and restored the regular list behavior for the GPU field.

lukasfrank (approved):

PR #773, opened 2026-09-02
Fix 727 network plugin fails silently after certificate rotation fails

### LINKED ISSUE #727: Network Plugin Kubeconfig Rotation fails silently
**Describe the bug**
If the network plugin kubeconfig rotation fails, the machine pool cannot progress in creating machines. This is due to the kubeconfig rotation being dispatched in the `Init` function of the network interface plugin (https://github.com/ironcore-dev/libvirt-provider/blob/cc7de6fc6fa69a481dd685b5fe3d5eb376aa8e91/internal/plugins/networkinterface/apinet/apinet.go#L88) without any health or readiness check.

The contract of the network interface plugin is also not quite well-formed here: We should research how e.g. k8s with the network interface plugin does both authentication and watching to make failures more visible and to be able to react on them. Example would be Calico to look at, as it also does api server communication.

friegger:
Errors yielded by the clients config controller during cert rotation are now hooked up to the health checks.

Fixes #727 

## Summary by CodeRabbit

* **New Features**
  * Health status endpoints now include configured network interface checks.
  * Health checks report **200 OK** only when libvirt and all configured checks pass; otherwise, they return **503 Service Unavailable** and identify the failing check.
* **Bug Fixes**
  * Provider startup failures are now logged at the normal error level, making them visible without increased verbosity.
* **Tests**
  * Added coverage for healthy states, libvirt disconnections, and individual or partial health-check failures.

### DIFF
--- a/cmd/libvirt-provider/app/app.go
+++ b/cmd/libvirt-provider/app/app.go
@@ -53,6 +53,7 @@ import (
 	"github.com/spf13/pflag"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
+	"k8s.io/apiserver/pkg/server/healthz"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
@@ -400,6 +401,9 @@ func Run(ctx context.Context, opts Options) error {
 	healthCheck := healthcheck.HealthCheck{
 		Libvirt: libvirt,
 		Log:     log.WithName("health-check"),
+	}
+	if nicConfigCtrl != nil {
+		healthCheck.Checkers = append(healthCheck.Checkers, healthz.NamedCheck(nicConfigCtrl.Name(), nicConfigCtrl.Check))
 	}
 
 	setupLog.Info("Starting resource claimer")
--- a/internal/healthcheck/health_check.go
+++ b/internal/healthcheck/health_check.go
@@ -6,21 +6,40 @@ package healthcheck
 import (
 	"net/http"
 
-	"github.com/digitalocean/go-libvirt"
 	"github.com/go-logr/logr"
 	libvirtutils "github.com/ironcore-dev/libvirt-provider/internal/libvirt/utils"
+	"k8s.io/apiserver/pkg/server/healthz"
 )
 
 type HealthCheck struct {
-	Libvirt *libvirt.Libvirt
+	Libvirt libvirtutils.Connector
 	Log     logr.Logger
+
+	// Checkers are additional named health checks that must all pass for the
+	// handler to report healthy (e.g. the apinet kubeconfig rotator).
+	Checkers []healthz.HealthChecker
 }
 
 func (h HealthCheck) HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	if err := libvirtutils.IsConnected(h.Libvirt); err == nil {
+	healthy := true
+	if err := libvirtutils.IsConnected(h.Libvirt); err != nil {
+		h.Log.Error(err, "failed to get active connection to libvirtd")
+		healthy = false
+	}
+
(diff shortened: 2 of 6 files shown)


internal/plugins/networkinterface/apinet/apinet.go
```diff
@@ -350,6 +361,27 @@ func (p *Plugin) startWatcher(ctx context.Context) error {
...
+			log.Error(err, "Informer cache stopped with error")
+			select {
+			case errCh <- err:
+			default:
+			}
+		}
+	}()
+
+	return errCh
+}
```
lukasfrank: this looks super hacky + I'm not sure if it's even working. If I'm not mistaken c.Start only errors when the context closes or on init

cmd/libvirt-provider/app/app.go
```diff
@@ -422,6 +426,25 @@ func Run(ctx context.Context, opts Options) error {
 		})
 	}
 
+	// Supervise the NIC plugin's informer watch (if it exposes cache errors), so a
+	// permanently-dead watch exits the process instead of silently stalling machines.
+	// A structural interface assertion (rather than a method on networkinterface.Plugin)
+	// keeps the plugin contract untouched: only apinet talks to a K8s API, so adding a
+	// health method to the interface would be scaffolding for the non-K8s plugins.
+	if cacheErrPlugin, ok := nicPlugin.(interface{ CacheErrors() <-chan error }); ok {
```
lukasfrank: I also don't really like the anonymous interface assertion because it's super error prone because we lose all compile time checks and then fail silently at runtime

lukasfrank (changes requested):

internal/plugins/networkinterface/apinet/apinet.go
friegger: Was just an experiment, removed.

cmd/libvirt-provider/app/app.go
friegger: Removed.

cmd/libvirt-provider/app/app.go
```diff
@@ -401,6 +402,9 @@ func Run(ctx context.Context, opts Options) error {
 		Libvirt: libvirt,
 		Log:     log.WithName("health-check"),
 	}
+	if nicConfigCtrl != nil {
+		healthCheck.Checkers = append(healthCheck.Checkers, healthz.NamedCheck(nicConfigCtrl.Name(), nicConfigCtrl.Check))
```
lukasfrank, line 406:
```suggestion
  healthCheck.Checkers = append(healthCheck.Checkers, nicConfigCtrl)
```
I think the config.Controller already fulfills  the checker

internal/healthcheck/health_check.go
```diff
@@ -6,21 +6,40 @@ package healthcheck
...
+			h.Log.Error(err, "health check failed", "check", checker.Name())
+			healthy = false
+		}
+	}
+
+	if healthy {
 		w.WriteHeader(http.StatusOK)
 	} else {
-		h.Log.Error(err, "failed to get active connection to libvirtd")
 		w.WriteHeader(http.StatusServiceUnavailable)
```
lukasfrank, line 43: we just return the header, wdyt about returning the failing checker?

lukasfrank (changes requested):

cmd/libvirt-provider/app/app.go
friegger: Yes, good point. I also changed the HealthCheck contract to handle checkers internally as private field. The libvirt connection test is now also just a named healthz check.

internal/healthcheck/health_check.go
friegger: I assume by returning you meant writing the failing check as plain text repsonse into the body. The assumption is now implemented.

lukasfrank (approved):

PR #777, opened 2026-09-07
Fix: size limit with disk image set

### LINKED ISSUE #776: Disk size not being honored with source image file set
**Describe the bug**
When a local disk vol is created with both image and size set, the size is not being taken into account. 

At the moment image and size are treated as mutually exclusive. While CreateOptions carries both fields, the implementation picks only the one or the other. Requests for larger volumes e.g. 10GB are ignored for  boot disks. 

`internal/raw/raw_exec.go` makes the same assumption:
```go
if o.SourceFile == "" {
    ... // size set
} else {
    ... // copy path, Size ignored
}
```

**To Reproduce**
Create a machine with a local disk volume specifying both an image and a size:

```go
Volumes: []*api.VolumeSpec{
    {
        Name:   "disk-0",
        Device: "oda",
        LocalDisk: &api.LocalDiskSpec{
            Image: &image,   // ghcr.io/ironcore-dev/gardenlinux/gardener:2150.4.0-kvm
            Size:  10 * 1024 * 1024 * 1024, // 10 GiB
        },
    },
},
```

**Expected behavior**
- Disks are created with the requested size in all scenarios. Image is written at the front and the file is expanded.
- A size smaller than the source image should be rejected.

**Additional context**

In my particular case, creating a disk with the gardenlinux image produces a 1.79GiB disk. The guest sees the same size as well.

This issue makes it impossible to set the size of a boot disk. There is even code specifically excluding boot disks from online resize.

`internal/controllers/machine_controller_volumes.go`:
```go
isBootDisk := desiredVolume.LocalDisk != nil && desiredVolume.LocalDisk.Image != nil
if lastVolumeSize != 0 && volumeSize != lastVolumeSize && !isBootDisk {
    // ...attacher.ResizeVolume()
}
```

Guests have only whatever free space the image happens to ship with. The disk fills quickly under any use, and there is no workaround.

brsmnv:
A local disk volume that sets both `image` and `size` ignored the size. This resulted in a disk at the size of the image rootfs while the requested size got ignored.. See the issue for the full write-up.

# Proposed Changes

- localdisk.Apply passes both options when image and size are set.
- Exec.Create extends the file to the requested size with `os.Truncate`.
- Guardrails and tests (cannot request undersized volumes or truncate the image)

Leaving `Size` unset preserves the current behaviour and the rootfs is matched.

Fixes #776 

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring local disk sizes when creating disks from boot images.
  * Disk images are expanded to the requested size when applicable.
  * Empty disks receive a default size when no size is specified.
  * Source image size is preserved when no larger size is requested.

* **Bug Fixes**
  * Prevented creation when the requested size is smaller than the source image.
  * Added validation for missing, zero, or negative disk sizes.
  * Failed disk creation no longer leaves behind incomplete disk files.


internal/raw/raw_exec_test.go
```diff
@@ -0,0 +1,112 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
friegger, line 1: The license headers in this project have meanwhile changed and omit the year.

internal/raw/raw_exec_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	sourceSize     int64 = 4096
+	grownSize      int64 = sourceSize * 4
+	emptyDiskSize  int64 = sourceSize * 2
+	undersizedSize int64 = sourceSize / 2
+)
+
+func TestExecCreateDiskSize(t *testing.T) {
```
friegger, line 21: The rest of the project uses Ginkgo tests, please use Gingko here as well for consistency.

friegger (changes requested):

internal/raw/raw_exec_test.go
brsmnv: Done

internal/raw/raw_exec_test.go
brsmnv: Fixed, didnt think of using ginkgo for non-controller tests but it reads better now.

brsmnv:
All comments and findings should be resolved now.  @friegger 

Please do check `internal/server/machine_internal_test.go` as I decided against Ginkgo here. Let me know if that's okay.

Going with plain `testing` means that the tests can run independently from the integration suite (which requires libvirt). The related change wasnt absolutely necessary, but I think it helps reject invalid requests early on. Helps with misleading status conditions downstream.

internal/raw/raw_exec.go
```diff
@@ -17,11 +17,24 @@ type Exec struct{}
...
+func (Exec) Create(filename string, opts ...CreateOption) (err error) {
 	o := &CreateOptions{}
 	o.ApplyOptions(opts)
 	log := ctrl.Log.WithName("raw-disk").WithValues("filename", filename)
 
+	defer func() {
+		if err != nil {
+			os.Remove(filename)
+		}
+	}()
```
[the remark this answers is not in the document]
brsmnv: Confirmed with a test, the suggested solution with the temp path is nonsense. On it.

internal/server/machine_internal_test.go
```diff
@@ -0,0 +1,61 @@
...
+		LocalDisk: &iri.LocalDisk{SizeBytes: -1},
+	})
+
+	if err == nil {
+		t.Fatalf("accepted a negative disk size, expected error for SizeBytes=%d", spec.LocalDisk.Size)
+	}
+
+	if !errors.Is(err, ErrInvalidRequest) {
+		t.Errorf("expected %v got %v for invalid negative request", ErrInvalidRequest, err)
+	}
```
friegger: NIT: We could still use the Ginkgo assertions like:
```
g := NewWithT(t)
srv := serverWithLocalDiskPlugin(t)
...
g.Expect(err).To(MatchError(ErrInvalidRequest))
```

friegger (approved):
Thanks for the contribution!

